PR #18, opened 2025-06-06
More modality stuff

samtoth:
Fixes the definition of modality
Proves some stuff about arrow maps and orthogonality conditions
Defines pullback-power and characterises the fibres of this as lifts of squares


samtoth:
TODO: 
 [ ] - Rename terminology for pullback squares as cartesian squares - to differentiate from cones that are pullbacks

PR #20, opened 2025-06-13
Defn: Flat modality

samtoth:
Some basic facts including:

 - Sigma closed
 - Defn of crisp induction
 - crisp induction implies:
   - lex
   - commutes with pb

Todo:
 - [x] Add changelog 
 - [x] Write intro to modal type theory


samtoth:
See https://github.com/samtoth/agda-synthetic-categories/tree/experimental/Interval for experiments with sharp modality

src/modalities/Flat/Flat.lagda.tree
```diff
@@ -0,0 +1,391 @@
...
+  qiso ._≅_.fwd-iso .snd .snd (mod♭ x) = refl
+}
+%```
+
+\p{If we have that #{\flat} is a lex modality, then (1),(2) and (3) are
+    all equivalent.}
+
+%```agda
+\agda{
+♭-pb3 : ∀ {@♭ 𝓤 𝓥 𝓦} {@♭ A : Type 𝓤} {@♭ B : Type 𝓥}
```
samtoth, line 376: Better name needed

PR #72, opened 2026-03-20
Characterisation of the covariant action of Hom

fredrik-bakke:
...and as a corollary associativity of composition.


fredrik-bakke:
Yeah, I'm having mysterious issues when building the website (but no errors). Gonna just hope it works as expected 😇

samtoth (commented):
Very good! Not many changes needed. 

 - And trees need to be renamed before merging

PR #69, opened 2026-03-12
Quasicoherence

markrd-williams:
Build up Lattice theory enough to do quasicoherence and show .

- [x] Reorganise algebras to be primarily bundled.
- [x] Give the universal property of a quotient algebras.
- [x] Define quasicoherent algebras.
- [x] Prove basic facts about quasicoherence.

PR #77, opened 2026-03-24
Exponentiability of the right endpoint inclusion

fredrik-bakke:
Defines exponentiability and shows that the right endpoint inclusion is exponentiable.

Builds on #63.


fredrik-bakke:
Also, I know this partly clashes with #69, I already mentioned this briefly to @markrd-williams.

PR #90, opened 2026-04-07
Fully faithful maps

fredrik-bakke:
I figured we'd need to populate the list of definitions of different orthogonality classes before we can get to the more meaty theorems. This PR also adds some infrastructure for right orthogonality classes.

Builds nonessentially on #87


src/Core/OrthogonalClosure.lagda.tree
```diff
@@ -171,11 +231,45 @@ right-orthogonal-left-cancel
...
+      {A : Type 𝓤} {B : Type 𝓥} {g h : A → B}
+    → g ~ h
+    → f ⊥ g
+    → f ⊥ h
+right-orthogonal-homotopy f {A = A} {B = B} {g = g} {h = h} H
+  = right-orthogonal←equiv' f
+      (mk-amap (id: A) (id: B) H)
+      (id-square-is-equiv {f = g})
+
+right-orthogonal-∘~
```
samtoth, line 252: I don't know if this is an idiom you will like (😬) but in the 3-for-2 lemma for equivalences for example, I write `is-equiv~∘`. I'm willing to accept this notation is better, but it should probabably be consistent

src/Core/OrthogonalClosure.lagda.tree
```diff
@@ -718,3 +812,117 @@ opaque
...
+
+%```agda
+\agda{
+module RightOrthogonalClassClosure
+  {𝓘 𝓤 𝓥} {I : Type 𝓘}
+  {A : I → Type 𝓤} {B : I → Type 𝓥}
+  (f : (i : I) → A i → B i)
+  where
+
+  ∈R
```
samtoth, line 847: Maybe this can be restated using the notion of `Global classes of maps`

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,448 @@
...
+}
+%```
+
+\subtree[stt-00G3]{
+\taxon{Definition}
+\title{Fully faithful maps in terms of the action on morphisms}
+\date{2026-03-11}
+\author{samueltoth}
+
+\p{A map #{F : A \to B} is called \em{fully faithful} if its action on synthetic
```
samtoth:
```suggestion
\p{A map #{F : A \to B} is called \em{fully faithful} if its [action on synthetic morphisms](stt-00CE)
```

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,448 @@
...
+}
+
+\subtree[fre-4004]{
+\title{Equivalence of definitions of fully faithful maps}
+\taxon{Theorem}
+\author{fredrikbakke}
+\date{2026-04-05}
+
+\p{
+The orthogonality-based and action-on-homs based predicates are equivalent.
```
samtoth: Links here

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,448 @@
...
+\proof{
+\p{
+Let #{\iota : \partial\Delta^1 \hookrightarrow \Delta^1} be the boundary inclusion.
+Given a map #{f : A \to B}, then we have an equivalence of arrows
+
+\quiver{
+\begin{tikzcd}
+(\Delta^1 \to A) && \Sigma_{(x\ y : A)\Hom_A(x,y)} \\
+\\
+\Hom(\iota , f) && \Sigma_{(x\ y : A)\Hom_B(f(x),f(y))}
```
samtoth: It took me a bit of time to reakuse that this `Hom` in the bottom left corner of this square was a map of arrows - since it clashes with the Hom as in synthetic morphisms. Maybe we could use `Arr` instead? I.e. the arrow category of Type, I'm guessing you used Hom here becaise it was what was used elsewhere?

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,448 @@
...
+    : is-ff f ⇔ is-cartesian (src×tgt-square f)
+  is-ff⇔is-cartesian-src×tgt-square
+    = ( is-ff→is-cartesian-src×tgt-square
+      , is-ff←is-cartesian-src×tgt-square)
+}
+%```
+}
+
+
+\subtree[fre-16AF]{
```
samtoth: Nice :) I'm a fan of having some non-formalised bits here and there, especially with a nice and consise proof. I could imagine such trees enticing someone to make a pull request

samtoth (commented):
(Oh and needs a changelog)

src/Core/OrthogonalClosure.lagda.tree
fredrik-bakke: `is-equiv-~∘` is better, because it has a clear dual `is-equiv-∘~` (i.e. with the homotopy pointing the other way) and I would not like `is-equiv∘~`. We could consider using a left arrow instead of a hyphen if you prefer: `is-equiv←~∘` and `is-equiv←∘~`

src/Synthetic/Categories/FullyFaithful.lagda.tree
fredrik-bakke: How about `Hom_{Arr}(ι, f)`? I find `Arr(ι, f)` slightly confusing since we'd also use the notation `Arr(𝒞)` for the arrow category of the category `𝒞`

src/Core/OrthogonalClosure.lagda.tree
fredrik-bakke: I guess `is-equiv∘~` is fine...

fredrik-bakke:
Comments are addressed @samtoth.

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,456 @@
...
+    → Σ[ x ∶ A ] Σ[ y ∶ A ] Hom B (f x) (f y)
+  arr-map-∂Δ¹-incl→ΣΣHom F
+    = ( F .Arrow-map.top ∂Δ¹-0
+      , F .Arrow-map.top ∂Δ¹-1
+      , mk-hom
+          (F .Arrow-map.bot)
+          (F .Arrow-map.comm ∂Δ¹-0)
+          (F .Arrow-map.comm ∂Δ¹-1))
+
+  ΣΣHom→arr-map-∂Δ¹-incl
```
samtoth:
```suggestion
  arr-map-∂Δ¹-incl←ΣΣHom
```

samtoth: And other places in this module where the arrow direction is wrong way round

src/Synthetic/Categories/FullyFaithful.lagda.tree
```diff
@@ -0,0 +1,456 @@
...
+
+\p{
+This result is analogous to \citet{Proposition 2.6.3}{Hau25}, although the definition
+of fully faithful maps and the method of proof are different.
+}
+
+\proof{
+\p{
+By \ref{stt-00LX}, this source-target square is equivalent to the square
+#{\mathrm{Exp}(\partial\Delta^1\text{-}\mathrm{incl}, f)}, and that square is
```
samtoth: Is this `Exp` meant to be the pullback power?

samtoth (commented):
Just spotted a few things when having last check through

src/Synthetic/Categories/FullyFaithful.lagda.tree
fredrik-bakke: This is referring to the square itself, not the pullback power, which is an arrow

fredrik-bakke: Replaced the notation with a diagram

samtoth: Yeah, maybe we could just mention that we will use the terminology `Exp` for the square in the pullback power module

fredrik-bakke: I actually think it is more useful to have the explicit diagram in this case. If need be later we can introduce that notation again :)

PR #92, opened 2026-04-08
Another funny segal condition

fredrik-bakke:
Considers another funny segal condition: locality at the inclusion $\Lambda^2_1 \to ({\Delta^1})^{\Delta^1}$. I show that such types are Segal via a retract argument, and that $\Delta^1$-discrete types are local at this inclusion via a cone contraction argument.

I have not yet determined a connection between this and left Segalness.

Builds on #86.


fredrik-bakke:
Closing this. I'll pull the useful changes to a different branch at some point.

PR #120, opened 2026-04-23
Auto rename action

### LINKED ISSUE #112: Integrate Tree ID renaming into merging workflow
It would be very useful if we could integrate the process of renaming trees into the merging process so that a maintainer may handle it non-locally and without the contributor's input.

Related: #113

samtoth:
Adds an action that can be manually triggered that auto renames trees in a PR


samtoth:
Closes #112

PR #122, opened 2026-04-28
Benchmarks workflow

fredrik-bakke:
Adds a benchmarking CI and a dedicated page for benchmarking <https://samtoth.github.io/agda-synthetic-categories/benchmarks/>, like this one: https://fredrik-bakke.github.io/agda-unimath/benchmarks/

The json/html/python files are taken verbatim from my agda-unimath fork, the workflow is adapted from the same fork, but it uses nix directly. 

I will have to run the workflow on the main branch to see that it works properly.

### DIFF
--- a/.codespell/ignore.txt
+++ b/.codespell/ignore.txt
@@ -13,3 +13,4 @@ falg
 bloc
 anull
 whitelist
+master
--- a/.codespell/requirements.txt
+++ b/.codespell/requirements.txt
@@ -1,0 +1,1 @@
+codespell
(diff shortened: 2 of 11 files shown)


samtoth:
Cool! Is there an option to change the scale from miliseconds to seconds?

fredrik-bakke:
No. I guess I can make it convert the unit though. Do you want the number rounded to some fixed point?

samtoth:
> No. I guess I can make it convert the unit though. Do you want the number rounded to some fixed point?

Seconds and 2 decimal places seems reasonable

PR #125, opened 2026-04-29
add `.gitattributes`

fredrik-bakke:
Makes it so literate agda files are reported as agda code in the repo, and trees are reported as text (since forester isn't supported by linguist).


samtoth:
Great!

PR #133, opened 2026-04-30
Change precedences for infix operators

fredrik-bakke:
Follows  agda-unimath's guidelines. Also changes the names of operators on homotopies, equivalences, quasi-equivalences, to use letter identifiers rather than relational symbol identifiers, making the distinction between operators and relations clearer. I also add a bare bones contributing doc to document that we use these guidelines.


fredrik-bakke:
Personally I think `~refl` actually looks prettier than `hrefl`, so I'm not convinced about that one.

samtoth:
> Personally I think `~refl` actually looks prettier than `hrefl`, so I'm not convinced about that one.

I agree with this, I think that one can be reverted, otherwise it seems to be an improvement. 

I also noticed there are some decimal precedences (which I didn't know you could do!) I wonder if we should just make some extra space in the single digit precedences. I don't think we need to stick exactly to unimath's table - I think we have more reasoning type combinators maybe, hence the need for the decimals

fredrik-bakke:
Done now.

> I also noticed there are some decimal precedences (which I didn't know you could do!) I wonder if we should just make some extra space in the single digit precedences. I don't think we need to stick exactly to unimath's table - I think we have more reasoning type combinators maybe, hence the need for the decimals

Agree with this, though it makes the agda-unimath guidelines less useful as a reference. This only entailed bumping pairing to have precedence level 4, however.

samtoth:
> Agree with this, though it makes the agda-unimath guidelines less useful as a reference. This only entailed bumping pairing to have precedence level 4, however.

If we diverge too far in the future, we can distribute our own modified guidelines, for now I think this is enough.

It has also occurd to me that we have `∙` and inverse for iff as well. I suppose we could just spell out `∙iff`?

CONTRIBUTING.md
samtoth, line 1: I think this would be better as a tree!

fredrik-bakke: `CONTRIBUTING.md` is specially recognized by github, which is where I was thinking this would be most useful

CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
+# Contributing
+
+> 🚧 This document is currently under construction. 🚧
+
+- We follow
+  [the agda-unimath guidelines for mixfix operators](https://unimath.github.io/agda-unimath/MIXFIX-OPERATORS.html).
```
samtoth: Would you be able to document the difference from the agda-unimath scheme. I think it is just the low/ single digit precedences?

samtoth (approved):

CONTRIBUTING.md
fredrik-bakke: Sure! let me just try and recall first 😅😅😅

fredrik-bakke: It turned out that the special cases were mistakes. We do have some extensions of the table though, which are now documented.

fredrik-bakke: Documenting the extensions led to a few further edits, sorry about that.

PR #137, opened 2026-05-01
Change -ap- name convention

markrd-williams:
just a quick one changing `Σ-ap-≃` and `Π-ap-≃` and related to `Σ-≃` and `Π-≃`


src/Core/PushoutSumsFamilies.lagda.tree
```diff
@@ -125,19 +125,19 @@ module _ {𝓤1 𝓤2 𝓤3 𝓤4 𝓤5 𝓤6 : Level}
...
     : (fp : base-Σ̂ᶠ F G)
-    → fam-Σ̂ᶠ F G fp ≃ fam-Σ̂ᶠ F H (Σ̂ᶠ-ap-≃ᵈᶠ-base ._≃_.fwd fp)
-  Σ̂ᶠ-ap-≃ᵈᶠ-family (a , x)
+    → fam-Σ̂ᶠ F G fp ≃ fam-Σ̂ᶠ F H (Σ̂ᶠ-≃ᵈᶠ-base ._≃_.fwd fp)
+  Σ̂ᶠ-≃ᵈᶠ-family (a , x)
     = *-≃ id≃ (e a .snd x)
 
-  Σ̂ᶠ-ap-≃ᵈᶠ : Σ̂ᶠ F G ≃ᶠ Σ̂ᶠ F H
-  Σ̂ᶠ-ap-≃ᵈᶠ = Σ̂ᶠ-ap-≃ᵈᶠ-base , Σ̂ᶠ-ap-≃ᵈᶠ-family
+  Σ̂ᶠ-≃ᵈᶠ : Σ̂ᶠ F G ≃ᶠ Σ̂ᶠ F H
```
fredrik-bakke: I suppose `Σ̂ᶠ-≃ᶠ` is more consistent with the other names

samtoth:
This is good. In TotalEquiv.lagda.tree we define `map-\Sigma` which maps both equivalences at once. I think `Σ-≃` would be better reserved for that one. Maybe what you now call `Σ-≃` could be `total-≃`- and then we could add `Σ-≃` as a new definition in TotalEquiv.lagda.tree?

src/Foundations/FibrewiseEquivalences.lagda.tree
```diff
@@ -52,10 +52,10 @@ is-total-equiv←is-fibrewise-equiv {f = f} feq = is-equiv←qinv qinv where
   qinv .snd .fst (a , b) = Σ-path→ (refl , A.retraction-fwd a .snd b)
   qinv .snd .snd (a , b) = Σ-path→ (refl , A.section-fwd a .snd b)
 
-Σ-ap-≃
+Σ-≃
   : ∀ {𝓤 𝓥 𝓦} {A : Type 𝓤} {B : A → Type 𝓥} {B' : A → Type 𝓦}
     → ((a : A) → B a ≃ B' a) → Σ A B ≃ Σ A B'
-Σ-ap-≃ eq = mk≃ (total-map (_≃_.fwd ∘ eq))
+Σ-≃ eq = mk≃ (total-map (_≃_.fwd ∘ eq))
                 (is-total-equiv←is-fibrewise-equiv (_≃_.fwd-is-equiv ∘ eq))
```
fredrik-bakke: In my opinion, `Σ-≃` should be reserved for the function that accepts both an equivalence in the base and an equivalence in the family over the base equivalence. The present definition might more accurately be described as `Σ-snd-≃` or `Σ-fam-≃`

src/Foundations/IsEquivIsProp.lagda.tree
```diff
@@ -108,19 +108,35 @@ section-coherence f (g , ε) = Σ (g ∘ f ~ id) λ η → f ◂ η ~ ε ▸ f
...
+    module A x = is-equiv (feq x)
+
+    qinv : quasi-inv (Π-map f)
+    qinv .fst = Π-map A.bwd
+    qinv .snd .fst =
+      λ g → WithFunExtω.funext→ FE λ a → A.η a (g a)
+    qinv .snd .snd =
+      λ g → WithFunExtω.funext→ FE λ a → A.ε a (g a)
+
+Π-≃
```
fredrik-bakke: Same here. This might more accurately be described as `Π-fam-≃` so the name `Π-≃` is 
reserved for the version that accepts a reparametrisation of the base too.

src/Foundations/FibrewiseEquivalences.lagda.tree
samtoth: Snap!

src/Foundations/SigmaProperties.lagda.tree
```diff
@@ -48,16 +48,16 @@ total-is-equiv←fst-is-equiv {B = B} {f} feq = is-equiv←qinv qusi where
   qusi .snd .fst (a , b) = Σ-path→ (η a ,  lem)
   qusi .snd .snd (a , b) = Σ-path→ (ε a ,  tr∘tr-sym' (ε a) b)
 
-Σ-ap-≃-fst
+Σ-≃-fst
```
fredrik-bakke: This reads like "the equivalence on `fst`" not "an equivalence from an equivalence on `fst`". The appropriate ordering seems to me to name this `Σ-fst-≃`

fredrik-bakke: Curious for @samtoth's thoughts on the naming of this one

src/Foundations/IsEquivIsProp.lagda.tree
markrd-williams: I see - my thought process is that the reparametrisation of the base would be `Π₁-≃` or some other : to me `Π-≃` reads as product of equivalences. Happy to be wrong. 

Also happy to go with `Π-map-≃` and `total-≃` since those are the names of the functions in question, although I think `total-≃` might affect proof readability.

src/Foundations/FibrewiseEquivalences.lagda.tree
fredrik-bakke: or `total≃`!

src/Foundations/SigmaProperties.lagda.tree
samtoth: Yeah if it is uniform for `fst` and `snd` I think that is best

src/Foundations/IsEquivIsProp.lagda.tree
markrd-williams: My main goal is to remove `-ap-` from these so, as the types don't contain the function `ap`. So any renaming works for me.

PR #143, opened 2026-05-16
Some closure properties of flat discrete types

fredrik-bakke:
Shows they're closed under
- [x] Dependent sums
- [x] cartesian products
- [x] The empty type
- [x] The unit type
- [x] Singletons
- [x] Retracts
- [x] Equivalences
- [x] Booleans
- [x] Natural numbers
- [x] Coproducts
- [x] Identity types


src/Foundations/CoproductFunctoriality.lagda.tree
```diff
@@ -0,0 +1,127 @@
...
+\author{fredrikbakke}
+
+\agda{
+infixl 10 _⨄ᶠ_
+_⨄ᶠ_
+  : ∀ {𝓘 𝓙 𝓤}
+      {I : Type 𝓘} {J : Type 𝓙}
+      (A : I → Type 𝓤) (B : J → Type 𝓤)
+    → I ⊎ J → Type 𝓤
+(A ⨄ᶠ C) (inl i) = A i
```
samtoth: This function is the recursion principal for `_⊎_` at `Type 𝓤`. I think it is correct that we have a bespoke name for this case, but since these maps are defined by pattern matching, it wont be definitionally equal to the recursor. I think it may as well be made definitionally equal (by defining it in terms of the recursor [⊎[_]](https://samtoth.github.io/agda-synthetic-categories/stt-002Y/))?

src/Foundations/CoproductFunctoriality.lagda.tree
```diff
@@ -0,0 +1,127 @@
...
+  {C : J → Type 𝓤} {D : J → Type 𝓥}
+  (f : (i : I) → A i → B i)
+  (g : ((j : J) → C j → D j))
+  where
+
+  ⨄₁ᶠ-carrier : Type (𝓘 ⊔ 𝓙 ⊔ 𝓤 ⊔ 𝓥)
+  ⨄₁ᶠ-carrier = (k : I ⊎ J) → (A ⨄ᶠ C) k → (B ⨄ᶠ D) k
+
+  infixl 10 _⨄₁ᶠ_
+  _⨄₁ᶠ_ : ⨄₁ᶠ-carrier
```
samtoth: I think this reads better with the type inlined

```suggestion
  _⨄₁ᶠ_ : (k : I ⊎ J) → (A ⨄ᶠ C) k → (B ⨄ᶠ D) k
```

src/Modalities/Cohesion/Connectivity.lagda.tree
samtoth, line 1: Maybe this module could be renamed, since it is mainly about `FlatDiscreteTypes` or alternatively could be split up into two modules, about flat discreteness and detecting connectivity

src/Modalities/Cohesion/Continuity.lagda.tree
```diff
@@ -263,7 +263,7 @@ large-family-separates←detects-continuity {I = I} G gc f f♭eq fsep
     rsquare-is-equiv .snd = global-transpose-is-equiv
 
     lemma : is-equiv (♭-map (postcomp ⟨♭| G i ⟩ (Arrow.f f)))
-    lemma = is-equiv←Arrow-equiv⁻¹ {F = rsquare} rsquare-is-equiv
+    lemma = src-is-equiv←Arrow-equiv (rsquare , rsquare-is-equiv)
```
samtoth:
```suggestion
    lemma 
      = src-is-equiv←Arrow-equiv 
         (rsquare , rsquare-is-equiv)
         (♭-map-is-equiv
```

src/Modalities/Cohesion/Flat.lagda.tree
```diff
@@ -232,16 +261,35 @@ _ = ♭-ind _ (ap♭ mod♭)
...
+♭-retract (i , r , H)
+  = ♭-map i
+  , ♭-map r
+  , λ where (mod♭ a) → ap♭ mod♭ (H a)
+}
+}
+
 \subtree[stt-0053]{
-\title{Flat commutes with #{Σ} and #{1}}
+\title{Flat commutes with  #{1}, #{Σ}, #{×} and #{⊎}}
```
samtoth: Now there is more closure properties in this tree, it might be best to at least split out `Sigma` into its own tree

src/Synthetic/Categories/Spaces.lagda.tree
```diff
@@ -80,8 +80,11 @@ opaque
     : ∀ {@♭ 𝓤 𝓥} (@♭ fa : Arrow 𝓤 𝓥) (let @♭ f = Arrow.f fa)
       → @♭ ((∀ {@♭ n} → is-equiv (♭-map (postcomp (□^ n) f))))
       → is-equiv f
-  cubes-separate' fa eq = cubes-separate fa
-                     ((is-equiv←Arrow-equiv {F = amap} amap-is-equiv (eq {0}))) eq where
+  cubes-separate' fa eq
+    = cubes-separate
+        (fa)
+        (tgt-is-equiv←Arrow-equiv (amap , amap-is-equiv) (eq {0}))
+        (eq) where
```
samtoth: I believe this is what is most inline with the rest of the codebase

```suggestion
  cubes-separate' fa eq
    = cubes-separate
        fa
        (tgt-is-equiv←Arrow-equiv (amap , amap-is-equiv) (eq {0}))
        eq where
```

samtoth (commented):
Very good - only some tiny nitpicks

Hold off on the tree rename as I will try and do it automatically

src/Foundations/CoproductFunctoriality.lagda.tree
fredrik-bakke: I completely agree, but having it written as it is currently serves a purpose. Using the entry  `⨄₁ᶠ-carrier` makes it easier to apply `⨄₁ᶠ` elsewhere when we construct intersections of localisations, and using it here also guarantees that `⨄₁ᶠ-carrier` is defined correctly.

samtoth: Hmm I'm not sure I fully understand your point. By 'makes it easier to apply `⨄₁ᶠ` elsewhere' do you mean makes it easier for the type checker or for the author?
Anyway I have two suggestions then: 

1) add a line 
```
_ : ⨄₁ᶠ-carrier
_ = _⨄₁ᶠ_
``` 
as a sort of 'unit test' if you will.

2) (which I think is probably more principled) add a definition somewhere for `fibrewise-map (A ⨄ᶠ C) (B ⨄ᶠ D)` since we also have definitions like `is-fibrewise-equiv` around the place. Does that still make it easy to apply `⨄₁ᶠ` in other places?

src/Modalities/Cohesion/Flat.lagda.tree
fredrik-bakke: I factored out coproducts, since that's what I added

src/Foundations/CoproductFunctoriality.lagda.tree
fredrik-bakke: It makes it easier for the author; we don't really care about the type of the family, we just want to define the coproduct of two localising families.

fredrik-bakke: Where should I put `fibrewise-map`?

samtoth: I think `Foundations.Functions` is probably reasonable

fredrik-bakke: your second suggestion makes most sense to me. I'll incorporate that.

fredrik-bakke:
done

samtoth:
Okay so I haven't been able to figure out the autorenamer. The issue is that the workflow is ran on your branch on your repo so the permissions are different - e.g. the commenting isn't working. I think it works fine for when the PR originates from agda-synthetic-categories - but otherwise it is still broken. 

I don't have anymore time today to look at this, so at some point would you be able to manually rename trees?

PR #148, opened 2026-05-19
Enable `--auto-inline` flag

fredrik-bakke:
This reduces typechecking time and memory usage by approximately 11% on my computer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ EVERYTHING_FILE ?= src/Everything.agda
 WATCH_DIR ?= src
 PORT ?= 1313
 DUP_DIR ?= ./trees/
-AGDA_FLAGS ?= --without-K --rewriting --guardedness --flat-split --level-universe --postfix-projections --local-confluence-check --no-qualified-instances -WnoWithoutKFlagPrimEraseEquality
+AGDA_FLAGS ?= --without-K --auto-inline --rewriting --guardedness --flat-split --level-universe --postfix-projections --local-confluence-check --no-qualified-instances -WnoWithoutKFlagPrimEraseEquality
 EVERYTHING_INPUTS := $(shell find src -type f \( -name '*.agda' -o -name '*.lagda.tree' \) ! -name 'Everything.agda' | sort)
 
 .PHONY: help generate-everything prepare-agda-datadir sync-forest-src typecheck benchmark-typecheck build-forest watch-agda check-port watch-forest server check-dup clean-agda clean-forester clean serve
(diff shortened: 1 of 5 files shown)


samtoth:
Okay that's pretty great results! I have read about what auto-inline does, I have some questions:

1. Are the extra implicit args you added needed because of auto-inline: If you mark `coe-cov-id` as NOINLINE, can Agda infer these arguments again?
2. Does it make goal display noticably worse/affect it at all? Reading a bit further, my impression is that it doesn't affect it.
3. Should we think about a hueristic of when to mark things `{-# NOINLINE #-}`, for example if marking `coe-cov-id` improves the implicit situation then what is it about that function that requires it?

fredrik-bakke:
Good questions. It's a bit weird that `coe-cov` is the only issue. Looking at its implementation:

```agda
coe-cov : ∀ {𝓤} {B : Δ¹ → Type 𝓤} {i j} → is-covariant B → i ≤ j → B i → B j
coe-cov {_} {B} {i} {j} Bc p = tr-cov-hom Bc (Δ¹-hom←≤ p)

opaque
  coe-cov-id : ∀ {𝓤} {B : Δ¹ → Type 𝓤} (bc : is-covariant B)
                 {i} (p : i ≤ i) → coe-cov {B = B} bc p ~ id
  coe-cov-id {B = B} bc {i} p x = happly (ap (tr-cov-hom bc) (Δ¹-hom-≤-refl p)) x
                              ∙ tr-cov-idH B bc x

  coe-cov-＝ : ∀ {𝓤} {B : Δ¹ → Type 𝓤} (bc : is-covariant B)
                 {i j} (p : i ＝ j) (q : i ≤ j) → coe-cov {B = B} bc q ~ tr B p
  coe-cov-＝ bc refl q = coe-cov-id bc _

  coe-cov-0≤1 : ∀ {𝓤} {B : Δ¹ → Type 𝓤} (bc : is-covariant B)
                  (p : i0 ≤ i1) → coe-cov {B = B} bc p ＝ tr-cov bc id
  coe-cov-0≤1 {B = B} bc p = ap (tr-cov-hom bc) (Δ¹-hom-0≤1 p)
```

It seems that it's sort of a coincidence that Agda was able to infer `B` in `Spaces`. Therefore I'd be happier with just having it accepting `B` explicitly.

samtoth:
> It seems that it's sort of a coincidence that Agda was able to infer `B` in Spaces. Therefore I'd be happier with just having it accepting `B` explicitly.

Yep, `is-covariant` isn't good at helping Agda infer. It could be fixed with making it a record type down the line, but for now better to be explicit. It's annoying that one has to use these 1 field records to get inference to work, I tried experimenting with the INJECTIVE_FOR_INFERENCE pragma but it didn't seem to help.

Since the results are good, I will merge this change. Maybe it will become clearer when (if ever) one should use NO_INLINE down the line.

Issue #150, opened 2026-05-20
Define algebraic factorisation systems


samtoth:
Should also refactor the prose in [[Axioms.FreeAlgebraicInjectives]](https://samtoth.github.io/agda-synthetic-categories/Axioms.FreeAlgebraicInjectives) to use more standard terminology.

Should also talk about the connection between algebraic factorisation systems and orthogonal factorisation systems

PR #151, opened 2026-05-20
chore: Fix weak orthogonator terminology

fredrik-bakke:
Replaces the terminology "weak orthogonator" with the established term "lifting operation", and disambiguates it from the term "lifting property".

#150


samtoth:
R.e. the naming "algebraically injective", I found the following remark from RSS:

> (In fact, JF (X) is the “free algebraically F -injective type on X”, cf. [[Bou18]](https://arxiv.org/abs/1811.09532).)

When I address #150 I can change the name to fit into that scheme or maybe remark why/where this name comes from

PR #156, opened 2026-05-21
Add pre-commit

fredrik-bakke:


samtoth:
Okay, I've had a read of the various enabled pre-comit modules and they all seem sensible. 

So let me just clarrify how this is intended to work, hopefully you can let me know i've got the right idea -- I have never used pre-comit before:

The pre-comit action doesn't change anything, but it will just fail if any of the checks don't pass. The developer is then encouraged to run pre-comit before each commit (I guess via git-hooks), and that will actually perform some changes e.g. to layouts and stuff. 

At the moment the pre-comit modules are mainly targetting the non-Agda parts of the repo. I think autoformatting these sorts of things is very sensible since we don't want to have to care about them very much. I still have some reservations about auto-formatting Agda code (with some obvious exceptions such as import sorting), but we can talk about that down the line!

fredrik-bakke:
> Okay, I've had a read of the various enabled pre-comit modules and they all seem sensible.
> 
> So let me just clarrify how this is intended to work, hopefully you can let me know i've got the right idea -- I have never used pre-comit before:
> 
> The pre-comit action doesn't change anything, but it will just fail if any of the checks don't pass.

No, most actions change the file contents, but if you prefer them not to and just point out suggested changes instead you can use `pre-commit --dry-run`.

> The developer is then encouraged to run pre-comit before each commit (I guess via git-hooks), and that will actually perform some changes e.g. to layouts and stuff.

I don't know about this, it's not how I use it, but it sounds reasonable. Maybe that is an alternative workflow.

> At the moment the pre-comit modules are mainly targetting the non-Agda parts of the repo. I think autoformatting these sorts of things is very sensible since we don't want to have to care about them very much. I still have some reservations about auto-formatting Agda code (with some obvious exceptions such as import sorting), but we can talk about that down the line!

For now I unfortunately don't have the resources to spend to implement any non-rudimentary Agda formatting. What I have in mind is simply inserting and removing some obvious line breaks. E.g. line breaks after a module declaration, and removing linebreaks if there are more than 3 in a row.

samtoth:
> > The pre-comit action doesn't change anything, but it will just fail if any of the checks don't pass.
> 
> No, most actions change the file contents, but if you prefer them not to and just point out suggested changes instead you can use `pre-commit --dry-run`.

Sorry, I shoud have specified better here, when I said 'action' here I meant the github action pre-commit.yaml. This doesn't seem to push any changes is what I was getting at/asking?

fredrik-bakke:
Ah, yes, that's correct. It just complains

README.md
```diff
@@ -17,6 +17,8 @@ the project and set up a server you can run `make server` from the nix shell, wi
 optional `PORT=` parameter (default: `1313`). To see all `make` targets run
 `make help`.
 
+<!-- TODO: installation instructions for pre-commit -->
```
samtoth: Should add a very minimal explanation here, even if it's just a link to the pre-commit documentation page

fredrik-bakke: ah yes my bad I forgot

fredrik-bakke:
@samtoth I'm not entirely sure how to best add pre-commit to the nix get-up. Maybe you could help me? Here's the agda-unimath flake for reference

https://github.com/UniMath/agda-unimath/blob/master/flake.nix

samtoth:
> @samtoth I'm not entirely sure how to best add pre-commit to the nix get-up. Maybe you could help me?

Since we are not using some specific version of Python3 it should be more simple than in that snippet. In shell.nix add the following:

```diff
    buildInputs = [
      (forester-server.packages.${system}.default)
      (pkgs.python3.withPackages (py-pkgs: []))
+     (pkgs.pre-commit)
      (pkgs.codespell)
      (pkgs.ripgrep)
    ] ++ drv.buildInputs;
```

(I tested this on my system and it works fine)

nix/agda-forester/default.nix
fredrik-bakke, line 1: @samtoth BTW whatever script generates these files should be updated to insert a final newline.

README.md
fredrik-bakke: Done

nix/agda-forester/default.nix
samtoth: Hmm I don't think it's possible to control this - it uses the external tool nix-thunk. Would it be possible to add these to an ignore list?

fredrik-bakke: but they should be in the repo?

samtoth: Yes - this way most nix users don't need nix-thunk installed - unless they want to change a dependency.

samtoth:
It looks like there are still changes to the nix/*.nix files as part of this PR - I think they should be unchanged right?

fredrik-bakke:
My bad! I thought I had applied the revert correctly but forgot to check

.prettierrc.json
fredrik-bakke, line 1: This is just taken verbatim from agda-unimath btw, with line width increased to 85.

samtoth: This just runs for these types of files right:

```
[markdown, yaml, javascript, json, html, shell]
```
?

fredrik-bakke: yes

fredrik-bakke:
> > @samtoth I'm not entirely sure how to best add pre-commit to the nix get-up. Maybe you could help me?
> 
> Since we are not using some specific version of Python3 it should be more simple than in that snippet. In shell.nix add the following:
> 
> ```diff
>     buildInputs = [
>       (forester-server.packages.${system}.default)
>       (pkgs.python3.withPackages (py-pkgs: []))
> +     (pkgs.pre-commit)
>       (pkgs.codespell)
>       (pkgs.ripgrep)
>     ] ++ drv.buildInputs;
> ```
> 
> (I tested this on my system and it works fine)

I wonder if it is cleaner to install pre-commit as a python package since the nix install already adds python. Thoughts? If so how do I do that precisely?

.prettierrc.json
samtoth: I'm happy with that then. The standards seem reasonable - and we hopefully wont spend much of our time looking at these file types

samtoth:
> I wonder if it is cleaner to install pre-commit as a python package since the nix install already adds python. Thoughts? If so how do I do that precisely?

Tbh the recommended way with nix seems to be to install it as a "top level" package. I suspect the end result of installing it as a python package vs like this would be the same anyway.

PR #160, opened 2026-05-26
Chore: Speed up with new version of agda-forester

### LINKED ISSUE #121: The "backlinks" and "related" sections are overcrowded by Agda code links
This creates a lot of noise, which makes these sections not very useful, for one. Additionally, on some pages it's even becoming a rendering issue, where the page takes a long time to load. E.g., @samtoth pointed to https://samtoth.github.io/agda-synthetic-categories/identity-def/index.xml.

~A suggested solution is to make the `\agda` macro open an anonymous subtree.~

samtoth:
Bumps the version of agda-forester to include the option of removing "forester native" links for the generated code. This speeds up the building of the forest quite significantly, and drastically reduces the overall size of the website from over 4.5GB down to around 350MB.

It also means that links from Agda code no longer register in the backlinks/related section - although links written in text will still appear there. This is both a positive and a negative, those sections were somewhat over crowded, so even though we lose some information from this change, the important links will at least be more obvious 

 - Updates the relavant uses of agda-forster in Makefile and nix-build script to use the new disable-backlinks option
 - Makes the nix-build command use the new `-j` multithreading option

Will merge when the upstream PR gets merged


nix/agda-forester/github.json
```diff
@@ -1,8 +1,8 @@
 {
   "owner": "samtoth",
   "repo": "agda-forester",
-  "branch": "main",
+  "branch": "htmlLinks",
```
fredrik-bakke: Is changing branches intentional? I presume you want to point to `main` after he upstream PR is merged

fredrik-bakke (approved):

fredrik-bakke:
It is unfortunate that we have to disable *all* backlinks. Though I think backlinks to, e.g., segal types will be overcrowded still.

nix/agda-forester/github.json
samtoth: Yes - just wanted to double check that the website output was all good before merging upstream

samtoth:
> It is unfortunate that we have to disable _all_ backlinks. Though I think backlinks to, e.g., segal types will be overcrowded still.

(Just in case there was any confusion - the backlinks for links in text still work)

But yeah I agree that this isn't the perfect solution. It's hard to think of a heuristic that would get the best of both worlds though. We can try to be better at adding more links in prose in the future - but as we were saying before - it would be nice to have infrastructure to make that process more painless

fredrik-bakke:
> (Just in case there was any confusion - the backlinks for links in text still work)

This was my precise confusion, based on your pr message

samtoth:
> This was my precise confusion, based on your pr message

Ah okay, I have updated to make this clearer

PR #163, opened 2026-05-28
The standard finite types

fredrik-bakke:
As promised, here's a small pull request defining the standard finite types as an indexed(?) inductive type family.


fredrik-bakke:
I would've liked to prove that `Fin` forms a preunivalent family in this PR, but I cannot be assed to clean up the mess that is `Modalities.Instances.Truncation` right now. #164

trees/macros/stt-macros.tree
```diff
@@ -29,6 +29,8 @@
 \def\false{\mathrm{false}}
 \def\Bool{\mathrm{Bool}}
 
+\def\Fin{\oeratorname{Fin}}
```
markrd-williams:
```suggestion
\def\Fin{\operatorname{Fin}}
```

fredrik-bakke: nice catch!

samtoth (approved):
Looks good.

Needs to be synced and trees-renamed

PR #169, opened 2026-05-29
Split off funext and pushouts from Axioms.UF

samtoth:
Resolves #140

Adds `Axiom.Funext` and tidies up `Axioms.UF`. Also updates some modules in Core where they should only rely on pushouts and funext.

Other misc cleanups

 - Makes naming around Funext a little more consistent
 - Defines TorsorialFamilies
 - Defines naive-funext and show equivalent to weak and full funext (co-authored by @fredrik-bakke)
 - Cleans up the Funext module a little

PR #170, opened 2026-05-29
The homotopy interval and function extensionality

fredrik-bakke:
Defines the homotopy interval and proves that function extensionality follows from its existence (because it has beta at the point constructors).


samtoth:
Merged into #169 now so closing this PR

Issue #114, opened 2026-04-22
Get rid of user-side `forester` dependency


fredrik-bakke:
It would be very beneficial if we could get rid of this dependency on the user side. This also includes removing the `forest-server` dependency.

samtoth:
I'm not sure what you mean with this, surely the user would like to see the forest as they are building it?

fredrik-bakke:
Can't they do that with the preview bot?

samtoth:
I suppose so, between commits. I guess it makes having agda-forester installed locally less important

samtoth:
I think maybe there is a difference here compared to how something like unimath works, where the code is the thing you write, and the website is generated from the code; in agda-synthetic-cats I would say the forest is the thing we are writing as much as the code is. So I don't see it as desirable to never build the forest locally. 

(Although I agree making it not 100% essential will help new contributors)

fredrik-bakke:
Over at agda-unimath we do not expect the user to build the website locally. That would considerably increase the workload of writing a contribution, and considerably raise the technical skill floor for contributors.

fredrik-bakke:
Ultimately, we should aim to make it as easy as possible to complete a PR. That includes one-shotting the final forest output itself IMO. By having the preview bot the contributor is reminded to check that final output.

samtoth:
> Ultimately, we should aim to make it as easy as possible to complete a PR. That includes one-shotting the final forest output itself IMO. By having the preview bot the contributor is reminded to check that final output.

I am in definite agreement with the first (and last) sentence, but I'm not sure about the second. In my mind this is comparable to saying that when writing, say in LaTeX, you would expect to one-shot the output. (Maybe this is your ideal state of writing LaTeX?).

Anyway, I agree with both of the subissues, which will have the result of limiting the requirement of having forester built locally.

fredrik-bakke:
> (Maybe this is your ideal state of writing LaTeX?)

No, I'm not saying I expect it to. I'm saying it should be a viable way to do it. To properly compare it to LaTeX we would want to have a similarly user-friendly IDE integration (e.g. with automatic building and previewing, and potentially error/warning reporting), and even then I would not say they are on the same footing as LaTeX is just a tool that is more broadly used.

[...]

PR #172, opened 2026-05-31
Updates renamer

### LINKED ISSUE #113: Improve tree ID renamer
The tree ID renamer script suffers from a series of issues which makes it impractical to work with, and that may be improved.

- Forester build dependency. Currently the script uses the build artifacts of forester, which is not great for multiple reasons:
  1. It requires that `forester` is installed, which we ultimately don't want. #114
  2. It requires that the project is built before the script can run, adding an extra step to the process of renaming IDs.
  3. This solution is fragile wrt. dirty repo states.
- Python dependency. It might be possible to implement the same functionality with `sed` (which ships with unix based systems (TODO: also document alternative for windows users(WSL?)))
- Every subtree must be manually ID'd. An alternative we've discussed is to autofill unID'd subtrees of the form `\subtree[]{` with the script (note the empty `[...]`).
- There seems to be performance issues with the current implementation where it takes multiple minutes to run.

### LINKED ISSUE #171: Makefile target for tree rename
A makefile target for tree renaming would at least rule out common errors such as the forest not being built/buildable

samtoth:
Some improvements to the tree renamer:

 - [x] Adds makefile target, `make list-trees` that uses rgrep and find, together with awk to list all the trees in the project.
 - [x] Updates the Python script to use this utility rather than calling `forester` to get a list of trees.
 - [x] Add a makefile target for renaming trees (#171):
   - Checks that there are no duplicate trees
   - Checks that the repo is synchronised with upstream/main
   - Checks that the repo is in a clean state
   - Displays the ouput a dry run, and asks for confirmation
   - Then performs the renaming
   - Optionally makes a git commit - enabled by default
 - [x] Add some contributor documentation for how to name trees and use the renamer etc.

This partially (or maybe completely?) addresses #113:

> The tree ID renamer script suffers from a series of issues which makes it impractical to work with, and that may be improved.

>   - Forester build dependency. Currently the script uses the build artifacts of forester, which is not great for multiple reasons:
>       1. It requires that `forester` is installed, which we ultimately don't want. [Get rid of user-side `forester` dependency #114](https://github.com/samtoth/agda-synthetic-categories/issues/114)
>       2. It requires that the project is built before the script can run, adding an extra step to the process of renaming IDs.
>       3. This solution is fragile wrt. dirty repo states.

I think this change fixes all of these issues, `forester` is no longer a dependency of the tree renamer, and the forest does not need to be build - or even buildable for this to work. 

>  - Python dependency. It might be possible to implement the same functionality with `sed` (which ships with unix based systems (TODO: also document alternative for windows users(WSL?)))

The Python dependency is still there. Maybe when I have time I can try and port the whole thing to bash, but since we have a dependency on Python now via `pre-commit` I don't think this is a top priority.

>  - Every subtree must be manually ID'd. An alternative we've discussed is to autofill unID'd subtrees of the form `\subtree[]{` with the script (note the empty `[...]`).

This is still an issue; however I think this could be solved by having tooling to automatically generate a new unique random ID in the editor for example. I don't think the proposed solution (of having blank subtrees) is ideal since we still want to be able to reference the subtrees before the final renaming.

>  - There seems to be performance issues with the current implementation where it takes multiple minutes to run.

Whilst I never encountered this with the last system - I suspect it was caused by forester itself having to build the forest before being able to generate the forest JSON data. This new system circumvents forester so this shouldn't be an issue.


fredrik-bakke:
> The Python dependency is still there. Maybe when I have time I can try and port the whole thing to bash, but since we have a dependency on Python now via pre-commit I don't think this is a top priority.

Completely agree—this is a bottom priority.

> This is still an issue; however I think this could be solved by having tooling to automatically generate a new unique random ID in the editor for example. I don't think the proposed solution (of having blank subtrees) is ideal since we still want to be able to reference the subtrees before the final renaming.

Let's write a separate issue for this feature request. To be clear, the suggestion is not to make it impossible to ID trees in advance, but to make it optional to ID them in advance.

> Whilst I never encountered this with the last system - I suspect it was caused by forester itself having to build the forest before being able to generate the forest JSON data. This new system circumvents forester so this shouldn't be an issue.

I suspect so too!

fredrik-bakke:
An incredibly useful thing for the make rename trees target would be for it to check that the branch is synchronised with the `main` branch.

Makefile
```diff
@@ -141,6 +143,53 @@ check-duplicate-trees:
...
+	} \
+	/\.tree$$/ { \
+		loc = $$0; \
+		id = $$0; \
+		sub(/^.*\//, "", id); \
+		sub(/\.tree$$/, "", id); \
+        	if (id ~ /^[a-zA-Z]{3}-[0-9A-Z]{4}/) { \
+		   printf "%s\n", id ;\
+		}\
+	}' | sort -r \
```
fredrik-bakke: maybe factor this to a separate script file?

Makefile
```diff
@@ -141,6 +143,53 @@ check-duplicate-trees:
...
+	else \
+	   echo "Dry running the renamer:"; \
+	   python3 scripts/rename_trees.py -n $(AUTHOR) src/ trees/  ; \
+	fi
+
+check-rename-trees:
+	@echo -n "Confirm changes? [y/N] " && read answer && [ $${answer:-N} = y ]
+
+rename-trees: check-sync-main rename-trees-dry check-rename-trees
+	python3 scripts/rename_trees.py $(AUTHOR) src/ trees/  ; \
```
fredrik-bakke: Surely this works?
```suggestion
	python3 scripts/rename_trees.py $(AUTHOR) src/ trees/
```

Makefile
```diff
@@ -141,6 +143,53 @@ check-duplicate-trees:
...
+
+check-sync-main:
+	@git fetch
+	@if !(git rev-list --left-right --count main...HEAD | head -c 1 | grep -q "^0$$") then \
+	    echo "You are out of sync with main"; \
+	    exit 1; \
+	fi
+
+
+
```
fredrik-bakke:
```suggestion

```

Makefile
```diff
@@ -10,7 +10,7 @@ DUP_DIR ?= ./trees/
 AGDA_FLAGS ?= --without-K --auto-inline --rewriting --guardedness --flat-split --level-universe --postfix-projections --local-confluence-check --no-qualified-instances -WnoWithoutKFlagPrimEraseEquality
 EVERYTHING_INPUTS := $(shell find src -type f \( -name '*.agda' -o -name '*.lagda.tree' \) ! -name 'Everything.agda' | sort)
 
-.PHONY: help generate-everything prepare-agda-datadir sync-forest-src typecheck benchmark-typecheck build-forest watch-agda check-port watch-forest server serve python-server check-duplicate-trees clean-agda clean-forester clean
+.PHONY: help generate-everything prepare-agda-datadir sync-forest-src typecheck benchmark-typecheck build-forest watch-agda check-port watch-forest server serve python-server check-duplicate-trees rename-trees clean-agda clean-forester clean
```
fredrik-bakke: All your added targets should be marked `.PHONY`

Makefile
fredrik-bakke: Or, I suppose not seeing as everything below follows the same pattern.

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,22 @@
+\title{Naming subtrees}
```
fredrik-bakke:
```suggestion
\title{ID'ing subtrees}
```
Not to be confused with titling subtrees.

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,22 @@
...
+
+
+\p{By default trees in the forest are assigned a unique sequential ID in the form
+\code{stt-XXXX}, where \code{XXXX} is a four digit base 36 number. To make
```
fredrik-bakke: Please elaborate on what base 36 looks like.

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,22 @@
...
+collaboration more simple, whilst you are working on a branch from which you intend
```
fredrik-bakke:
```suggestion
collaboration simpler, whilst you are working on a branch from which you intend
```

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,22 @@
...
+
+\p{There are two exceptions to this rule.}
+\ul{
+\li{Trees with the taxon \em{reference} should
+be named either using the initials of the authors and the year, or with a name
+if it is a reference to a software project.}
+
+\li{Trees with a taxon \em{person} should be given the name of the person all in
+lower case, no separator between words.}
+}
```
fredrik-bakke:
```suggestion
\p{There are two exceptions to this rule:
\ul{
\li{Trees with the taxon \em{reference} should
be named either using the initials of the authors and the year, or with a name
if it is a reference to a software project.}

\li{Trees with a taxon \em{person} should be given the name of the person all in
lower case, no separator between words.}
}
}
```

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,22 @@
...
+to submit a pull request, you may name your trees \code{abc-XXXX} where \code{abc}
+can be any lowercase ASCII string (such as the first three letters of your name),
+and \code{XXXX} is still a four digit base 36 number. This number still needs to
+be unique within your namespace; however, it need not be sequential.}
+
+\p{There are two exceptions to this rule.}
+\ul{
+\li{Trees with the taxon \em{reference} should
+be named either using the initials of the authors and the year, or with a name
+if it is a reference to a software project.}
```
fredrik-bakke: Our current references don't all follow this rule. Please update those references.

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,6 @@
+\title{Renaming trees}
+
+
+\p{Only once a pull request is ready to be merged should you rename
+the trees. This can be achieved by running \code{make rename-trees AUTHOR=abc},
+this script will also ensure that you are synced with the main branch.}
```
fredrik-bakke:
```suggestion
\p{
Only once you've been notified by a reviewer should you rename
your trees. This can be achieved by running \code{make rename-trees AUTHOR=abc}.
If you are not synchronised with the main branch this script will return an error.
}
```

CONTRIBUTING.md
```diff
@@ -32,3 +32,7 @@ with some extensions.
 | 3                | Parametric pairing operators: `_,_`, `_∷_`                                                                                                                        |
 | 0-1              | Reasoning syntaxes                                                                                                                                                |
 | -∞               | Function type formation: `_→_`                                                                                                                                    |
+
+## Naming trees
```
fredrik-bakke: ID'ing

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,6 @@
+\title{Renaming trees}
```
fredrik-bakke: Should we change the terminology to "ID'ing" or something else? The current terminology can at times be confused with titleing I think.

CONTRIBUTING.md
fredrik-bakke: Maybe

```suggestion
## Determining tree IDs
```

trees/stt/tot-0976.tree
samtoth: Yeah this is a good point. "Assigning canonical tree IDs" would make the most sense to me in written english. I could change the makefile target to `assign-canonical-ids`?

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,6 @@
...
+
+
+\p{Only once a pull request is ready to be merged should you rename
```
fredrik-bakke: Please also add a paragraph explaining what the purpose of tree identifiers is and why the contributor cannot just use `stt-` directly, and why we want it to be `stt-` on main.

fredrik-bakke (changes requested):
Glad to have this fix. As you anticipated I want some additions/changes to the documentation.

trees/stt/tot-0976.tree
fredrik-bakke: `assign-ids-trees` is better

fredrik-bakke:
The following identifiers are also currently in use and should be assigned new IDs:

```text
stt-2082
stt-0A2A
stt-0A2B
stt-0A2C
stt-0A2E
```

Makefile
```diff
@@ -141,6 +143,31 @@ check-duplicate-trees:
 		exit err ; \
 	}'
 
+list-trees:
+	 ./scripts/list-trees $(DUP_DIR)
+
+check-sync-main:
+	@git fetch
+	@if !(git rev-list --left-right --count main...HEAD | head -c 1 | grep -q "^0$$") then \
+	    echo "You are out of sync with main"; \
```
fredrik-bakke: Please also suggest the appropriate action to take in this scenario. I.e., merge upstream/main into your branch *and then try again

fredrik-bakke: Oh! Also check that the current state of the repo is clean. Renaming should never happen in a dirty state.

Makefile
```diff
@@ -141,6 +143,31 @@ check-duplicate-trees:
...
+	else \
+	   echo "Dry running the renamer:"; \
+	   python3 scripts/rename_trees.py -n $(AUTHOR) src/ trees/  ; \
+	fi
+
+confirm-assign-tree-ids:
+	@echo -n "Confirm changes? [y/N] " && read answer && [ $${answer:-N} = y ]
+
+assign-tree-ids: check-sync-main assign-tree-ids-dry confirm-assign-tree-ids
+	python3 scripts/assign_tree_ids.py $(AUTHOR) src/ trees/
```
fredrik-bakke: Finally, the script should make an immediate commit so to avoid the user adding other edits to the rename commit. (We do not want the user to edit it, because it makes it difficult to revert the renaming later should there be a conflict with a different PR).

fredrik-bakke: Maybe add a NOCOMMIT flag that can be used if people want it not to commit?

samtoth:
@fredrik-bakke I believe all the comments have now been addressed

trees/ref/forester.tree
```diff
@@ -0,0 +1,7 @@
+\title{Forester}
+\author{jonsterling}
+\meta{external}{www.forester-notes.org}
+
+\p{\em{Forester} is a tool for authoring, exploring, and sharing scientific and mathematical hypertexts. It is your lab notebook, your journal, your blackboard, and the home of your lecture notes.}
```
fredrik-bakke: line wrap

trees/stt/stt-00S2.tree
```diff
@@ -0,0 +1,33 @@
+\title{Tree IDs}
+
+\p{In [[forester]], each tree has an ID, given by the name of the file
```
fredrik-bakke, line 5:
```suggestion
\p{In [[Forester]], each tree has an ID, given by the name of the file
```

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,19 @@
...
+it is also checked as part of the CI.
+}
+
+\p{Only once you've been notified by a reviewer should you assign canonical IDs
+to your trees. This can be achieved by running \code{make assign-tree-ids
+AUTHOR=abc}. The repository should be in a clean state and synchronised with
+the upstream main branch. If either of these criteria are not met, this script
+will return an error. Note that it is important that this process only happens
+when the pull request is ready to be merged as notified by a maintainer, as
+ otherwise there will likely be issues with duplicate/conflicting tree IDs.
```
fredrik-bakke:
```suggestion
otherwise there will likely be issues with duplicate/conflicting tree IDs.
```

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,19 @@
...
+can be any lowercase ASCII string (such as the first three letters of your name),
+and \code{XXXX} is still a four digit base 36 number. This number still needs to
+be unique within your namespace; however, it need not be sequential – the makefile
+target \code{check-duplicate-trees} can be used to test this locally, and
+it is also checked as part of the CI.
+}
+
+\p{Only once you've been notified by a reviewer should you assign canonical IDs
+to your trees. This can be achieved by running \code{make assign-tree-ids
+AUTHOR=abc}. The repository should be in a clean state and synchronised with
```
fredrik-bakke, line 21: Does `\code` not preserve newlines?

trees/stt/stt-00S2.tree
samtoth: The tree reference is `forester` but it will display as `Forester` which is the title of that tree

CONTRIBUTING.md
```diff
@@ -32,3 +32,7 @@ with some extensions.
 | 3                | Parametric pairing operators: `_,_`, `_∷_`                                                                                                                        |
 | 0-1              | Reasoning syntaxes                                                                                                                                                |
 | -∞               | Function type formation: `_→_`                                                                                                                                    |
+
+## Determining tree IDs
```
fredrik-bakke:
```suggestion
## Assigning tree IDs
```

fredrik-bakke (approved):
Very nice!!

trees/stt/stt-00S2.tree
fredrik-bakke: oh my bad

fredrik-bakke:
Good job on the documentation, btw, it's in pristine condition now 💎

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,33 @@
...
+
+\li{Trees with a taxon \em{person}, living in \code{trees/people}, should be
+given the name of the person all in lower case, no separator between words.}
+}
+}
+
+\p{When multiple people are working on the project at the same time, it is
+infeasible to keep these IDs unique and sequentially ordered, and so we have
+a process to make assigning correct IDs as simple as possible.
+}
```
fredrik-bakke: This should go at the beginning of `tot-0976` actually

trees/stt/tot-0324.tree
```diff
@@ -0,0 +1,33 @@
...
+in the forest. For more information, please see the Forester documentation
+on [tree addresses](https://www.forester-notes.org/0073/index.xml). You
+may also find [this document](https://www.forester-notes.org/0052/index.xml)
+to be a useful introduction and overview of the Forester system.}
+
+
+\p{By default, trees in our forest are assigned a unique sequential ID in the form
+\code{stt-XXXX}, where \code{XXXX} is a four digit [base 36 number](https://www.forester-notes.org/0074/index.xml) – i.e. consisting of
+the alphabet \code{0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ}.
+}
```
fredrik-bakke: Move the following text to a new paragraph at the bottom of this tree:

> For more information about trees and forests, please see the Forester documentation
on [tree addresses](https://www.forester-notes.org/0073/index.xml). You may also find [this document](https://www.forester-notes.org/0052/index.xml) to be a useful introduction and overview of the Forester system.


And do
```suggestion
in the forest.}

\p{
By default, trees in our forest are assigned a unique sequential ID in the form
\code{stt-XXXX}, where \code{XXXX} is a four digit 
[base 36 number](https://www.forester-notes.org/0074/index.xml) – i.e. consisting 
of the alphabet \code{0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ}.
}
```

EDITED

fredrik-bakke: Actually. The last paragraph here should be merged with the next paragraph starting `\p{There are two exceptions to this rule:`

trees/stt/tot-0976.tree
samtoth: Nope :)

trees/stt/tot-0324.tree
samtoth: I have changed it to what I think you intended, could you double check it?

samtoth: Screenshot for reference

Screenshot 2026-06-04 at 14 57 28

fredrik-bakke: perfect!

trees/stt/tot-0976.tree
```diff
@@ -0,0 +1,26 @@
+\title{Assigning tree IDs}
+\author{samueltoth}
+\date{2026-06-04}
+
+\p{When multiple people are working on the project at the same time, it is
+infeasible to keep these IDs unique and sequentially ordered, and so we have
```
fredrik-bakke:
```suggestion
infeasible to keep the tree IDs unique and sequentially ordered, and so we have
```

fredrik-bakke (approved):

PR #175, opened 2026-06-02
Add logo to frontmatter

### LINKED ISSUE #174: Add logo
Since we have a logo now I think it's nice if we use it.

- [agda-logo.zip](https://github.com/user-attachments/files/28507879/agda-logo.zip)
- [logo-3-simplex.zip](https://github.com/user-attachments/files/28507882/logo-3-simplex.zip)

fredrik-bakke:
Here's what the index looks like with this change:

image

And here's what the readme looks like:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,10 @@ prepare-agda-datadir:
 	@chmod -R u+w "$(AGDA_DATADIR)/lib/prim" || true
 	@mkdir -p "$(AGDA_DATADIR)/lib/prim/_build"
 
+prepare-forest-assets:
+	@mkdir -p output/agda-synthetic-categories/assets
+	@cp assets/logo-wide-transparent.svg output/agda-synthetic-categories/assets/
+
 typecheck: $(EVERYTHING_FILE) prepare-agda-datadir
 	@mkdir -p "$(AGDA_DATADIR)" "$(AGDA_DATADIR)/lib"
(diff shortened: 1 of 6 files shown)


samtoth (dismissed):
Looks great!!

fredrik-bakke:
Maybe the website logo is a bit big?

samtoth:
> Maybe the website logo is a bit big?

Looks reasonable to me on the screenshots. I made a 'fix' to the preview bot a day or two ago that seemed to break it. I'm trying to fix the fix now, so let me get that up and running so that I can see it in situe - and check the CI runs properly

samtoth:
http://agda-synthetic-categories.toth.co.uk/preview/175/index/index.xml

I think the size looks right to me - at least when the window is large. The formatting gets a bit messed up on a small window/phone. I don't know how challenging it would be to fix that with inline css? Maybe it could just dissapear if the window is smaller than some size?

fredrik-bakke:
> http://agda-synthetic-categories.toth.co.uk/preview/175/index/index.xml 

I can't connect to this page on my phone for some reason.

samtoth:
> > http://agda-synthetic-categories.toth.co.uk/preview/175/index/index.xml
> 
> I can't connect to this page on my phone for some reason.

Hmm weird - works on my phone. And actually I have just checked and the styling doesn't look that bad on the phone... maybe fixing this isn't top priority right/best use of time right now

PR #158, opened 2026-05-22
Imports subtrees

fredrik-bakke:
Adds a new macro `\agda-imports` for a folded-by-default subtree with the imports of each module.
The agda imports block should not contain imports that rename entries, `Axiom` imports, or definitions.

The only agda files that do not have an agda imports block now are:

```text
src/Axioms/UF.lagda.tree
src/Foundations/Universes.lagda.tree
src/Foundations/Prelude.lagda.tree
src/Foundations/Reasoning.lagda.tree
```

Additionally,
- Minimizes the foundations imports that happen before the main module declaration, to make it easier to see what's needed for the module declaration itself.
- Removes the distributivity assumption on lattices where not needed, which is everywhere.
- Assumes a cohesive interval also in `Synthetic.Categories.UniversalElements`, like everywhere else.
- Removes some redundant imports and sorts the `Foundations.Prelude` file.

Builds on #156. Resolves #64.

PR #178, opened 2026-06-04
bug: Fix `\agda-imports` macro name

fredrik-bakke:
I had forgotten to rename the actual macro in #158, this fixes that.

As an aside, it's strange to me that forester does not pick up on this with an error.

### DIFF
--- a/trees/macros/basic-macros.tree
+++ b/trees/macros/basic-macros.tree
@@ -5,7 +5,7 @@
  \stopverb
 }
 
-\def\imports[~body]{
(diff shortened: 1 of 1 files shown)


samtoth:
> As an aside, it's strange to me that forester does not pick up on this with an error.

Yeah someday, when either of us has a bit more time, it's probably worth investing in some kind of Forester linter to pick up on these sort of things, and relatedly to pick up broken links. 

It might actually be the case that Forester outputs warnings for this rather than errors, and we could just check this in CI

PR #184, opened 2026-06-06
Edits to text for strictification

fredrik-bakke:


src/Synthetic/Categories/Strictification.lagda.tree
```diff
@@ -430,6 +431,20 @@ see their explanation for now:
       → P (Hom-comp-prestruct 𝐴)
     strictify P = tr P (sym Hom=Homʸ-comp-prestruct)
 }
+
+\p{
+The strictification operator state the following: to prove a property on the
```
samtoth:
```suggestion
The strictification operator states the following: to prove a property on the
```

src/Synthetic/Hom.lagda.tree
samtoth:
```suggestion
A homomorphism system should be analogous to an
[identity system](Foundations.IdentitySystems) on types, but for homomorphisms.
```

To me this reads a little better but feel free to ignore

samtoth (changes requested):

trees/stt/stt-00LM.tree
```diff
@@ -58,3 +38,5 @@ agree via the constructed equivalence.
 \li{\ref{Wild.Categories.Structure}}
 }
 }
+
+\transclude{Synthetic.Categories.Strictification}
```
samtoth: Sorry I missed this the first time, but I think it is not right to transclude an entire module as part of a changelog. Maybe just an important definition/theorem instead?

fredrik-bakke: I found it made a lot more sense than the previous version. The tree it transcludes is already as summarized as I want it.

samtoth: Hmm I don't just don't think it feels right to have an entire transcluded module. Generally the idea of the changelogs is to summarise and add a bit of human colouring that wouldn't feel right as part of the module, but I don't think transcluding an entire module really serves either of those purposes. I think the tree as you have written is fine, and the module can be linked to, as opposed to transcluded, and IMO this would be more in-keeping with the style of the rest of the changelogs.

samtoth (approved):

Issue #192, opened 2026-06-08
The family (i = 0) → X is covariant for a covariant family X


fredrik-bakke:
And more generally that covariant families are closed under exponentiating by contravariant families.

This should follow from the fact that left-anodyne maps are closed under base change by right-fibrations

PR #197, opened 2026-06-09
Prove the disjunctive property

rwbarton:
Proves the disjunctive property from
1. Types with initial objects separate
2. Directedness of the interval
3. The boundary of the interval covers the global points.


fredrik-bakke:
~~This is work towards #192~~

fredrik-bakke:
> (which follows from quasicoherence -- not formalized yet)

@nimarasekh is formalising this 👍 

EDIT: The formalisation is finished.

fredrik-bakke:
@rwbarton I hope you don't mind, I'm refactoring this PR to prove the disjunctive property instead of the conjunctive property, and we hope that the dual will be a direct (judgmental even) corollary with the right definitions. In addition, I'm generalising it to the conditions

1. The interval is directed (this is now established terminology in the library following your original suggestion).
2. The booleans cover the crisp elements of the interval.
3. Types with an initial object separate.

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
```diff
@@ -164,12 +164,12 @@ types-with-initial-object-separate'←finite-cubes-separate sep
...
-\p{If the inclusion #{∂Δ¹ \to \Delta^1} is [#{\flat}-connected](stt-0103), then
-so is the inclusion of points #{3 \to \Delta^2}.}
+\p{If the inclusion #{∂Δ¹ \hookrightarrow Δ¹} is [#{♭}-surjective](stt-0103), then
+so is the inclusion of points #{\Fin 3 \hookrightarrow Δ²}.}
 
 \proof{
-We exhibit the map #{3 \to \Delta^2} as a retract of the product of two copies
-of #{∂Δ¹ \to Δ¹}.
+We exhibit the map #{\Fin 3 \hookrightarrow Δ²} as a retract of the product of two
+copies of #{∂Δ¹ \hookrightarrow Δ¹}.
```
samtoth: These maps aren't embeddings unless we assume the interval is non-trivial. Which we don't require in this file.

src/Synthetic/DisjunctiveProperty.lagda.tree
```diff
@@ -0,0 +1,291 @@
+\title{The disjunctive property}
+\author{reidbarton}
+\author{fredrikbakke}
```
samtoth, line 3:
```suggestion
\author{fredrikbakke}
\import{stt-macros}
```

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
fredrik-bakke: Oh yeah. The boundary inclusion is tho

samtoth: The boundary as we have defined it isn't quite right without non-triviality. I.e. we have that it is equivalent to the booleans but it should be the type `(i : \II) x (i = 0 * i = 1)`. If it was defined as the latter then it would be an embedding.

I suppose I was just being pedantic tho, it seems reasonable to keep the hooked arrow since the intuition is that they are embeddings, and non-triviality is such a harmless assumption.

fredrik-bakke: > non-triviality is such a harmless assumption.

Agree to disagree

src/Synthetic/DisjunctiveProperty.lagda.tree
```diff
@@ -0,0 +1,292 @@
...
+##{f : A \to \sum_{i,j:\II} (i \lor j = 1)} lifts along the disjunctive inclusion.
+}
+
+\proof{
+\p{
+Write #{f(x) = ((i_x,j_x),p_x)}. To lift #{f}, it suffices to show
+#{(i_x = 1) \lor (j_x = 1)} for every #{x : A}. First observe that
+#{(i_a = 1) \lor (j_a = 1)}. Since #{f(a)} is crisp and the disjunctive inclusion is
+#{Δ¹}-surjective, it suffices to show this proposition when #{f(a)} comes from the
+domain of the inclusion, but there the disjunction holds by definition.
```
samtoth: It might be a bit of a skill issue, but I find this sentence really hard to parse.

> Since #{f(a)} is crisp and the disjunctive inclusion is
#{Δ¹}-surjective, it suffices to show this proposition when #{f(a)} comes from the
domain of the inclusion, but there the disjunction holds by definition.

src/Synthetic/DisjunctiveProperty.lagda.tree
```diff
@@ -0,0 +1,292 @@
...
+##{\iota : \Sigma_{(i,j : \II)}((i = 1) \lor (j = 1)) \hookrightarrow \Sigma_{(i,j : \II)}((i \lor j = 1))}
+is an equivalence. By the separation axiom, it suffices to check, for crisp
+types with initial objects #{A}, that
+##{\modflat{A \to \Sigma_{(i,j : \II)}((i = 1) \lor (j = 1))} \hookrightarrow \modflat{A \to \Sigma_{(i,j : \II)}(i \lor j = 1)}}
+is an equivalence. Since #{\iota} is an embedding it in turn suffices to show that
+every crisp map #{A \to \Sigma_{(i,j : \II)}((i \lor j = 1))} merely lifts along
+#{\iota}. Finally, by the assumption that the boundary inclusion of #{\II} is
+#{♭}-surjective #{\iota} is also #{♭}-surjective.
+Hence, the result follows by [disjunctivity propagating from initial objects](bar-648E).
+}
```
samtoth, line 255: Very nice :)

samtoth (approved):
Looks great

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
samtoth: Wait I'm confused now, they should either both be embeddings or neither – e.g. by the same retract argument as used show that they are both flat-surjective.

fredrik-bakke: Oh, right, The definition of ∂Δ¹ is bad! It should be `(i : Δ¹) × ((i = 0) ∨ (i = 1))`

src/Synthetic/DisjunctiveProperty.lagda.tree
fredrik-bakke: How do you find the current version?

fredrik-bakke: I'm going ahead and resolving this comment so I can merge.

Issue #208, opened 2026-06-11
Refactor lattices to come equipped with an inequality relation


fredrik-bakke:
This ensures that the opposite is well-behaved, since inequality will just be given by swapping order, which we use a lot.

PR #209, opened 2026-06-11
Conjunctive types

fredrik-bakke:
Dualises disjunctive types, using the opposite lattice. Note that in this PR we do not show simplicial types are conjunctive. This we postpone until #208 is implemented, which would make this proof trivial too.

### DIFF
--- a/src/Synthetic/Conjunctive.lagda.tree
+++ b/src/Synthetic/Conjunctive.lagda.tree
@@ -20,7 +20,7 @@ import Synthetic.Disjunctive
 module Conj = Synthetic.Disjunctive.Disj Δ¹ (Lattice-op I)
 }
 
-\subtree[stt-00TM]{
+\subtree[fre-3FEB]{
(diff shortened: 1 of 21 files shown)


src/Synthetic/Conjunctive.lagda.tree
```diff
@@ -0,0 +1,90 @@
...
+module Synthetic.Conjunctive {𝓘} (Δ¹ : Type 𝓘) (I : Lattice Δ¹) where
+
+open Lattice I renaming (0l to i0; 1l to i1)
+
+open import Algebra.Lattice.Instances.Opposite
+
+import Synthetic.Disjunctive
+module Conj = Synthetic.Disjunctive.Disj Δ¹ (Lattice-op I)
+}
+
```
samtoth, line 22: Maybe add an explanation that we reexport everything from disjunctive with the opposite lattice

samtoth (approved):

PR #213, opened 2026-06-11
Full subtypes

emilyriehl:
This gives a preliminary definition of full subtypes together with a proof that the first projection is fully faithful. Note that nothing here requires any types to be categories so perhaps this should be elsewhere? Note also full subtypes are not packaged as a record, so there is a bit of an asymmetry with the subtypes api.


src/Core/Orthogonal.lagda.tree
```diff
@@ -159,6 +160,13 @@ has-unique-extensions f P
   = ∀ (F : ArrFam-map f P)
     → is-singleton (Extᵈ f (P ∘ F .ArrFam-map.bot) (F .ArrFam-map.top))
 
+is-prop-Extᵈ-Subtype
+  : ∀ {𝓤 𝓥 𝓦 𝓜} {A : Type 𝓤} {B : Type 𝓥}
+      {C : Type 𝓦} (f : A → B)
+      (P : Subtype C 𝓜) → ∀ (F : ArrFam-map f (P .Subtype.family)) → is-prop (Extᵈ f (P .Subtype.family ∘ F .ArrFam-map.bot) (F .ArrFam-map.top))
```
samtoth:
```suggestion
  : ∀ {𝓤 𝓥 𝓦 𝓜} {A : Type 𝓤} {B : Type 𝓥}
      {C : Type 𝓦} (f : A → B)
      (P : Subtype C 𝓜) 
      (F : ArrFam-map f (P .Subtype.family))
    → is-prop (Extᵈ f (P .Subtype.family ∘ F .ArrFam-map.bot) (F .ArrFam-map.top))
```

src/Core/Orthogonal.lagda.tree
```diff
@@ -159,6 +160,13 @@ has-unique-extensions f P
...
+is-prop-Extᵈ-Subtype f P F = Σ-is-prop (Π-is-prop (P .Subtype.family-is-subtype ∘ F .ArrFam-map.bot)) λ a → Π-is-prop λ a'  → ＝-is-prop (((P .Subtype.family-is-subtype ∘ F .ArrFam-map.bot) ∘ f) a') _ _
```
samtoth: To keep to a line limit of 85 chars:

```suggestion
is-prop-Extᵈ-Subtype f P F
  = Σ-is-prop
      (Π-is-prop (P .Subtype.family-is-subtype ∘ F .ArrFam-map.bot))
      (λ a → Π-is-prop
        (λ a' → ＝-is-prop
                  (((P .Subtype.family-is-subtype ∘ F .ArrFam-map.bot) ∘ f) a') _ _))
```

src/Modalities/Cohesion/Sharp.lagda.tree
```diff
@@ -311,23 +288,48 @@ modext♯-is-equiv
...
+\subtree[rie-1423]{
+\title{Sharp preserves propositions}
+\author{emilyriehl}
+\taxon{Lemma}
+
+\p{The modality #{\sharp} preserves propositions.}
+
+\agda{
+♯-is-prop : ∀ {𝓤} {@♯ A : Type 𝓤} → @♯ is-prop A → is-prop ⟨♯| A ⟩
+♯-is-prop f x y = modext♯ (mod♯ (f ((x .counit♯)) ((y .counit♯))))
```
samtoth:
```suggestion
♯-is-prop f x y = modext♯ (mod♯ (f (x .counit♯) (y .counit♯)))
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+\date{2026-06-09}
+\author{emilyriehl}
+\import{stt-macros}
+
+\agda{
+open import Foundations.Universes
+open import Algebra.Lattice
+
+module Synthetic.Categories.FullSubcategories
+
```
samtoth:
```suggestion
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+  {@♭ 𝓘} (@♭ Δ¹ : Type 𝓘) (@♭ I : Lattice Δ¹)
+  where
+
+open Lattice I renaming (0l to i0; 1l to i1)
+}
+
+\agda-imports{
+\agda{
+
```
samtoth:
```suggestion
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+homotopy-∂Δ¹-incl ∂Δ¹-1 = refl
+homotopy-∂Δ¹-incl ∂Δ¹-0 = refl
+}
+
+\subtree[rie-1234]{
+\title{Full subtypes}
+
+\p{Recall that a [subtype](Foundations.Subtypes) of a type #{A} is a family of propositions #{φ} indexed by #{A}. Formally, this data is given by a family of types that we abusively denote #{φ : A \to \UU} together with a proof that each type #{φ a} is a proposition for all #{a : A}. From this data, one may construct a type #{Σ_{A} φ} with an embedding #{Σ_{A} φ \to A}.}
+
+\p{Given a crisp type #{A :^\flat \UU} and a family of propositions #{φ} indexed by #{^\flat A}, the \em{full-subtype} is defined to be #{Σ_{A} \overline{φ}} where #{\overline{φ}} denotes the adjoint transpose of the family #{φ} under the adjunction #{\flat ⊣ \sharp}.}
```
samtoth:
```suggestion
\p{Given a crisp type #{A :^\flat \UU} and a family of propositions #{φ} indexed by #{\flat A}, the \em{full-subtype} is defined to be #{Σ_{A} \overline{φ}} where #{\overline{φ}} denotes the adjoint transpose of the family #{φ} under the adjunction #{\flat ⊣ \sharp}.}
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+
+homotopy-∂Δ¹-incl : (ε♭: Δ¹ ∘ ∂Δ¹-to-♭Δ¹) ~ ∂Δ¹-incl
+homotopy-∂Δ¹-incl ∂Δ¹-1 = refl
+homotopy-∂Δ¹-incl ∂Δ¹-0 = refl
+}
+
+\subtree[rie-1234]{
+\title{Full subtypes}
+
+\p{Recall that a [subtype](Foundations.Subtypes) of a type #{A} is a family of propositions #{φ} indexed by #{A}. Formally, this data is given by a family of types that we abusively denote #{φ : A \to \UU} together with a proof that each type #{φ a} is a proposition for all #{a : A}. From this data, one may construct a type #{Σ_{A} φ} with an embedding #{Σ_{A} φ \to A}.}
```
samtoth: Can you line wrap these paragraphs

src/Synthetic/Categories/FullSubcategories.lagda.tree
samtoth: I think this definition should `#{Σ_{A} \sharp \overline{φ}}` where `\sharp` is the dependent sharp modal operator `\sharp Type -> Type`, since `\overline{φ}` has type `A \to \sharp Type`

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+  full-subtype = Σ A (λ a → ⟨♯| mod♭ a ∈ φ ⟩)
+
+  full-subtype-incl : full-subtype → A
+  full-subtype-incl = fst
+
+  full-subtype-family : (a : A) → Type 𝓥
+  full-subtype-family a = ⟨♯| mod♭ a ∈ φ ⟩
+
+  is-prop-full-subtype-family : (a : A) → is-prop (full-subtype-family a)
+  is-prop-full-subtype-family a = ♯-is-prop (∈-is-prop φ (mod♭ a))
```
samtoth: Our convention is to have things like `is-prop` or `is-equiv` as postfixes, so it reads a bit more like written English:


```suggestion
  full-subtype-family-is-prop : (a : A) → is-prop (full-subtype-family a)
  full-subtype-family-is-prop a = ♯-is-prop (∈-is-prop φ (mod♭ a))
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+
+\agda{
+  module _ (ax : is-equiv ∂Δ¹-to-♭Δ¹) where
+
+    is-ff-full-subtype-incl : is-ff (full-subtype-incl )
+    is-ff-full-subtype-incl _ =
+      left-orthogonal-homotopy _ homotopy-∂Δ¹-incl
+        (left-orthogonal-∘ (ε♭: Δ¹) ∂Δ¹-to-♭Δ¹ _
+          (are-orthogonal-♭counit-full-subtype-incl Δ¹) (equiv⊥ ax _))
+}
```
samtoth: This could be separated into its own tree, with a very brief informal proof

samtoth (changes requested):
I have addressed some simple codestyle issues. 

To address the question about where this should go, I agree that it is more general than just full subcategories and should probably go in `Modalities.Cohesion` somewhere. To me it kind of looks like the factorisation of a map into an sharp-equivalence and a sharp-codiscrete map, except that would apply if the input data was a type family over A - not over `Flat A` - so I'm still not 100% sure what the right name for this module would be.

To me either of the following options would be acceptable:

 1) Leaving the entire thing in FullSubcategories - with a remark added that this is actually a more general concept, which we can abstract later/when we need it.
 
 2) Creating a module in Modalities.Cohesion, maybe called something like sharp-modal factorisation - but maybe we can come up with a better name. And then moving all of the non-category specific stuff there.

Maybe @fredrik-bakke might have some opinions on naming, as you were involved in the conversation in Sweden about what the correct generalisation of this concept is.

fredrik-bakke:
> To address the question about where this should go, I agree that it is more general than just full subcategories and should probably go in `Modalities.Cohesion` somewhere. To me it kind of looks like the factorisation of a map into an sharp-equivalence and a sharp-codiscrete map, except that would apply if the input data was a type family over A - not over `Flat A` - so I'm still not 100% sure what the right name for this module would be.

I believe you meant to say the factorisation of a map into a sharp-connected map followed by a sharp-codiscrete map.

> To me either of the following options would be acceptable:
> 
>     1. Leaving the entire thing in FullSubcategories - with a remark added that this is actually a more general concept, which we can abstract later/when we need it.
> 
>     2. Creating a module in Modalities.Cohesion, maybe called something like sharp-modal factorisation - but maybe we can come up with a better name. And then moving all of the non-category specific stuff there.
> 
> 
> Maybe @fredrik-bakke might have some opinions on naming, as you were involved in the conversation in Sweden about what the correct generalisation of this concept is.


Is it the sharp image factorisation? If so I believe the correct place is `Modalities.Cohesion.SharpImage` Otherwise, let's put it in `FullSubcategories` for now and worry about it another time.

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,130 @@
...
+
+  is-prop-full-subtype-family : (a : A) → is-prop (full-subtype-family a)
+  is-prop-full-subtype-family a = ♯-is-prop (∈-is-prop φ (mod♭ a))
+
+  Subtype-full-subtype : Subtype A 𝓥
+  Subtype-full-subtype = mk-subtype is-prop-full-subtype-family
+
+  are-orthogonal-fam-̱♭counit-full-subtype-family :
+    {@♭ 𝓦 : Level} → (@♭ X : Type 𝓦) → are-orthogonal-fam (ε♭: X) full-subtype-family
+  are-orthogonal-fam-̱♭counit-full-subtype-family X =
```
samtoth:
```suggestion
  ε♭⊥full-subtype-family :
    ∀ {@♭ 𝓦} → (@♭ X : Type 𝓦) → are-orthogonal-fam (ε♭: X) full-subtype-family
  ε♭⊥full-subtype-family X =
```

samtoth:
> I believe you meant to say the factorisation of a map into a sharp-connected map followed by a sharp-codiscrete map.

Yes I suppose so, but since sharp is lex these are the same.

> Is it the sharp image factorisation? If so I believe the correct place is `Modalities.Cohesion.SharpImage` Otherwise, let's put it in `FullSubcategories` for now and worry about it another time.

It's not exactly -- it would be if the input data was a family over `A` rather that `Flat A`. But with `Flat A` I am not really clear on what it is. So I suppose `FullSubcategories` is best for now

fredrik-bakke:
Sounds good! 👍

src/Synthetic/Categories/FullSubcategories.lagda.tree
emilyriehl: Done.

src/Synthetic/Categories/FullSubcategories.lagda.tree
emilyriehl: Done.

src/Synthetic/Categories/FullSubcategories.lagda.tree
emilyriehl: Done.

src/Synthetic/Categories/FullSubcategories.lagda.tree
emilyriehl: I've edited the text as suggested. Let me know what you think.

src/Synthetic/Categories/FullSubcategories.lagda.tree
emilyriehl: I made this its own subtree called "Universal property of full subtypes" but included also the orthogonality with the flat counits. Note this means I copied some of the agda code from the FullSubtype module into a different subtree. Can modules span subtrees like this or should I close the first module and open a new one?

emilyriehl:
> Sounds good! 👍

My understanding of this discussion is that I should not move this file for now. 

Note for future reference something else that should be moved later, which is the homotopy relating the flat counit and the boundary inclusion for the 1-simplex.

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,157 @@
...
+}
+
+\subtree[rie-2718]{
+\title{Universal property of full subtypes}
+\author{emilyriehl}
+
+\p{The inclusion of a full subtype is orthogonal to all counits of the modality
+#{\sharp}. This orthogonality means that to define a map
+#{f : X \to Σ_{A} \overline{φ}} is to define a map #{f : \flat X → A} for which the
+elements #{f x} satisfy the predicate #{φ}.}
```
samtoth: I believe that this should be something like: ... is to define a map #{f : X \to A} which when restricted to crisp elements, f(x) satisfies φ.

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,157 @@
...
+∂Δ¹-to-♭Δ¹ ∂Δ¹-0 = mod♭ i0
+
+homotopy-∂Δ¹-incl : (ε♭: Δ¹ ∘ ∂Δ¹-to-♭Δ¹) ~ ∂Δ¹-incl
+homotopy-∂Δ¹-incl ∂Δ¹-1 = refl
+homotopy-∂Δ¹-incl ∂Δ¹-0 = refl
+}
+
+\subtree[rie-1234]{
+\title{Full subtypes}
+\author{emilyriehl}
```
samtoth, line 58:
```suggestion
\author{emilyriehl}
\taxon{Definition}
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,157 @@
...
+  full-subtype-family-is-prop a = ♯-is-prop (∈-is-prop φ (mod♭ a))
+
+  Subtype-full-subtype : Subtype A 𝓥
+  Subtype-full-subtype = mk-subtype full-subtype-family-is-prop
+}
+}
+
+\subtree[rie-2718]{
+\title{Universal property of full subtypes}
+\author{emilyriehl}
```
samtoth, line 97:
```suggestion
\author{emilyriehl}
\taxon{Lemma}
```

src/Synthetic/Categories/FullSubcategories.lagda.tree
samtoth: This is okay, since it is a named module

samtoth (approved):

src/Synthetic/Categories/FullSubcategories.lagda.tree
```diff
@@ -0,0 +1,157 @@
...
+
+\p{Recall that a [subtype](Foundations.Subtypes) of a type #{A} is a family of
+propositions #{φ} indexed by #{A}. Formally, this data is given by a family of types
+that we abusively denote #{φ : A \to \UU} together with a proof that each type #{φ a}
+is a proposition for all #{a : A}. From this data, one may construct a type
+#{Σ_{A} φ} with an embedding #{Σ_{A} φ \to A}.}
+
+\p{Given a crisp type #{A :^\flat \UU} and a family of propositions #{φ} indexed by
+#{\flat A}, the \em{full-subtype} is defined to be #{Σ_{A} \sharp \overline{φ}} where
+#{\overline{φ} : A \to \sharp UU} denotes the adjoint transpose of the family #{φ}
```
samtoth:
```suggestion
#{\overline{φ} : A \to \sharp \UU} denotes the adjoint transpose of the family #{φ}
```

samtoth:
Thanks for the contribution :)

PR #214, opened 2026-06-13
Add symmetry to ≤ for lattices

fredrik-bakke:
Redefines `≤` via

$$
  (i ≤ j) \coloneqq ((i \land j) = i) \times ((i \lor j) = j)
$$ 

Not entirely sure if this is the right step yet, but I wanted to try it out after it was suggested by @rwbarton.


fredrik-bakke:
This refactoring job looks to be a bit of a rabbit hole...

PR #221, opened 2026-06-25
I-algebra structure on simplices

ecavallo:
Contains proofs that
- Δ⁰ is a singleton
- Δ¹ is equivalent to the base lattice
- Δⁿ is an I-algebra via the diagonal map I → Δⁿ, and is distributive when I is a distributive lattice
- Δ¹ is an initial I-algebra

After defining the diagonal map I → Δⁿ, I redefined the top and bottom elements of Δⁿ using it. This required one change in another file.

### DIFF
--- a/src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
+++ b/src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
@@ -19,6 +19,7 @@ module Algebra.Lattice.Instances.Free.Inductive
 \agda-imports{
 \agda{
 open import Algebra.Poset
+open import Core.Coslices
 open import Modalities.Instances.Truncation
 open import Modalities.SeparatedTypes
 
@@ -65,18 +66,24 @@ last-Δ (x :≥: j) = j
 Δ-suc ._≊_.fwd-qinv .snd .fst (x :≥: j) = refl
 Δ-suc ._≊_.fwd-qinv .snd .snd = ~refl
 
+Δ-diagonal : ∀ {n} → I → Δ^ n
+Δ-diagonal-≤ : ∀ {i n} → i ≤ last-Δ {n} (Δ-diagonal i)
+
+Δ-diagonal {zero} i = ·
+Δ-diagonal {suc n} i = (Δ-diagonal i :≥: i) ⦃ Δ-diagonal-≤ {i} {n} ⦄
+
+Δ-diagonal-≤ {i} {zero} = auto!
+Δ-diagonal-≤ {i} {suc n} = auto!
+
 Δ-1 : ∀ {n} → Δ^ n
-Δ-1-≤ : ∀ {i n} → i ≤ last-Δ {n} (Δ-1)
-
-Δ-1 {zero} = ·
-Δ-1 {suc n} = (Δ-1 :≥: i1) ⦃ Δ-1-≤ {i1} {n} ⦄
-
-Δ-1-≤ {i} {zero} = auto!
-Δ-1-≤ {i} {suc n} = auto!
+Δ-1 = Δ-diagonal i1
 
 Δ-0 : ∀ {n} → Δ^ n
-Δ-0 {zero} = ·
-Δ-0 {suc n} = (Δ-0 :≥: i0)
+Δ-0 = Δ-diagonal i0
+
+Δ-1-≤ : ∀ {i n} → i ≤ last-Δ {n} (Δ-1)
+Δ-1-≤ {i} {zero} = auto!
+Δ-1-≤ {i} {suc n} = auto!
 
 Δ∗∗ : ℕ → Type∗∗ 𝓘
 Δ∗∗ n = mk-type∗∗ (Δ^ n) Δ-0 Δ-1
@@ -156,6 +163,38 @@ opaque
 instance
   Δ-Is-n-truncated : ∀ {n k} → Is-truncated (𝕋←ℕ k) (Δ^ n)
   Δ-Is-n-truncated = truncation-instance 0 Δ-is-0-truncated
+
+}
+}
+
+\subtree[uma-0000]{
+\title{#{Δ⁰} is a singleton}
+\taxon{Lemma}
+\date{2026-06-25}
+\author{ecavallo}
+
+\agda{
+Δ⁰-is-singleton : is-singleton (Δ^ 0)
+Δ⁰-is-singleton .centre = ·
+Δ⁰-is-singleton .central · = refl
+}
+}
+
+
+\subtree[uma-0001]{
+\title{#{Δ¹} is the base lattice}
+\taxon{Lemma}
+\date{2026-06-25}
+\author{ecavallo}
+
+\agda{
+Δ¹-diagonal-is-equiv : is-equiv (Δ-diagonal {1})
+Δ¹-diagonal-is-equiv = is-equiv←qinv qinv
+  where
+  qinv : quasi-inv (Δ-diagonal {1})
+  qinv .fst (· :≥: j) = j
+  qinv .snd .fst _ = refl
+  qinv .snd .snd (· :≥: j) = Δ-ext (· :≥: refl)
 }
 }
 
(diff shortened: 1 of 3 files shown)


src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -156,6 +163,38 @@ opaque
 instance
   Δ-Is-n-truncated : ∀ {n k} → Is-truncated (𝕋←ℕ k) (Δ^ n)
   Δ-Is-n-truncated = truncation-instance 0 Δ-is-0-truncated
+
```
fredrik-bakke:
```suggestion
```

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+}
+
+\agda{
+Δ-Lattice-is-lattice-algebra
+  : ∀ {n} → is-lattice-hom L (Δ-Lattice n) Δ-diagonal
+Δ-Lattice-is-lattice-algebra .is-lattice-hom.pres-∨ = Δ-ext Δ-diagonal-∨-path
+  where
+  Δ-diagonal-∨-path
+    : {n : ℕ} {j k : I}
+      → Δ-path {n} (Δ-diagonal (j ∨ k)) (Δ-diagonal j Δ-∨ Δ-diagonal k)
```
fredrik-bakke:
```suggestion
    : {n : ℕ} {j k : I}
    → Δ-path {n} (Δ-diagonal (j ∨ k)) (Δ-diagonal j Δ-∨ Δ-diagonal k)
```

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+  Δ-diagonal-∨-path {zero} = ·
+  Δ-diagonal-∨-path {suc n} = Δ-diagonal-∨-path :≥: refl
+Δ-Lattice-is-lattice-algebra .is-lattice-hom.pres-∧ = Δ-ext Δ-diagonal-∧-path
+  where
+  Δ-diagonal-∧-path
+    : {n : ℕ} {j k : I}
+      → Δ-path {n} (Δ-diagonal (j ∧ k)) (Δ-diagonal j Δ-∧ Δ-diagonal k)
```
fredrik-bakke:
```suggestion
    : {n : ℕ} {j k : I}
    → Δ-path {n} (Δ-diagonal (j ∧ k)) (Δ-diagonal j Δ-∧ Δ-diagonal k)
```

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+Δ-Lattice-is-lattice-algebra .is-lattice-hom.pres-1 = refl
+
+Δ-Algebra : ∀ n → Algebra L (Δ^ n)
+Δ-Algebra n .Algebra.str-Lattice = Δ-Lattice n
+Δ-Algebra n .Algebra.ι = Δ-diagonal
+Δ-Algebra n .Algebra.ι-alg = Δ-Lattice-is-lattice-algebra
+
+Δ-Algebra-is-distributive : ∀ {n}
+  → is-distributive L
+  → is-distributive-algebra (Δ-Algebra n)
```
fredrik-bakke:
```suggestion
Δ-Algebra-is-distributive 
  : ∀ {n}
    → is-distributive L
    → is-distributive-algebra (Δ-Algebra n)
```

src/Synthetic/Categories/NecklaceLifting.lagda.tree
```diff
@@ -63,7 +63,7 @@ Wedge-Δ-incl-0-1-is-equiv = mk-eqv
   ( ι₂
   , pushout-ind _
       (mk-coconeᵈ
-        (λ where · → sym (glue tt))
+        (λ where · → ap ι₂ (Δ-ext (· :≥: refl)) ∙ sym (glue tt))
```
fredrik-bakke, line 66: Did this have to be fixed because you changed the definitions of `Δ-1` and `Δ-0`? It'd be better if we could factor out the relevant lemmas (not sure what they should be right now) and put them in `Inductive` so that they can be used and issues like this don't happen downstream again.

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+  where
+  Δ-diagonal-∧-path
+    : {n : ℕ} {j k : I}
+      → Δ-path {n} (Δ-diagonal (j ∧ k)) (Δ-diagonal j Δ-∧ Δ-diagonal k)
+  Δ-diagonal-∧-path {zero} = ·
+  Δ-diagonal-∧-path {suc n} = Δ-diagonal-∧-path :≥: refl
+Δ-Lattice-is-lattice-algebra .is-lattice-hom.pres-0 = refl
+Δ-Lattice-is-lattice-algebra .is-lattice-hom.pres-1 = refl
+
+Δ-Algebra : ∀ n → Algebra L (Δ^ n)
```
fredrik-bakke, line 338: I'm open to be persuaded but it seems to me that `Δ-Algebra` and `Δ-Lattice` shouldn't be capitalised, i.e., they should be `Δ-algebra` and `Δ-lattice` respectively.

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+  Δ¹←0-is-equiv = Δ¹-diagonal-is-equiv
+
+  0←Δ¹-quasi-inv : LatticeAlgHom-quasi-inv Δ¹←0
+  0←Δ¹-quasi-inv = quasi-inv←is-LatticeAlgHom-equiv Δ¹←0-is-equiv
+
+  0←Δ¹ = 0←Δ¹-quasi-inv .fst
+
+  ret : ∀ φ → LatticeAlgHom-∘ (LatticeAlgHom-∘ φ Δ¹←0) 0←Δ¹ ＝ φ
+  ret φ
+    = ext! λ p → ap (λ ψ → φ .fst .fst (ψ .fst .fst p)) (0←Δ¹-quasi-inv .snd .snd)
```
fredrik-bakke: Forgive me if I'm wrong, but aren't you essentially directly applying the equivalence $I \simeq \Delta^1$ here? Could the proof be shortened using that? I'm envisioning an argument of the form: "This map has contractible fibres because it is an equivalence by the 3-for-2 property since L is initial and Δ¹ is L"

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -263,3 +302,80 @@ _Δ-∨_ {suc n} (p :≥: j) (p' :≥: j') = ((p Δ-∨ p') :≥: (j ∨ j')) 
...
+
```
fredrik-bakke:
```suggestion
```

fredrik-bakke (approved):

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
```diff
@@ -65,19 +66,25 @@ last-Δ (x :≥: j) = j
...
 Δ-suc ._≊_.fwd-qinv .snd .snd = ~refl
 
+Δ-diagonal : ∀ {n} → I → Δ^ n
+Δ-diagonal-≤ : ∀ {i n} → i ≤ last-Δ {n} (Δ-diagonal i)
+
+Δ-diagonal {zero} i = ·
+Δ-diagonal {suc n} i = (Δ-diagonal i :≥: i) ⦃ Δ-diagonal-≤ {i} {n} ⦄
+
+Δ-diagonal-≤ {i} {zero} = auto!
+Δ-diagonal-≤ {i} {suc n} = auto!
```
fredrik-bakke, line 77: When you make additions to a preexisting subtree as you've done here, don't forget to add yourself as a contributor by adding the line `\contributor{ecavallo}` some place close to the top of the declaration. If the contribution is large (guidelines pending), you can make yourself an author.

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
ecavallo: Fine with me, but I think I was following the pattern of `Free-Lattice` in `Algebra.Lattice.Instances.Polynomial`.

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
ecavallo: So far I've always aligned the arrow to the right of the colon (also elsewhere in this PR). Is there some reason why these are different?

fredrik-bakke: This indentation makes all the types align. When there's a forall operator, the types are shifted in one indentation level which is why it's different elsewhere.

src/Synthetic/Categories/NecklaceLifting.lagda.tree
ecavallo: The `Δ-ext (· :≥: refl)` here (also appearing elsewhere in the file) is used to equate two different terms that both look like `· :≥: i` but have different proofs that the chain is (trivially) monotone. Since the new definitions of `Δ-1` and `Δ-0` have a different proof there, this identity became necessary where it wasn't before.

Maybe it would be useful to have a special case of `Δ-ext` defined for this kind of thing. I'm not sure how to avoid accidentally taking advantage of coincidences where it isn't necessary, though.

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
ecavallo: Ah, makes sense, thanks!

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
ecavallo: What map do you mean in your envisioned argument? I agree that this comes down to $I \simeq \Delta^1$ (as an equivalence of $I$-algebras), but I didn't see a cleaner way to get to the result. On paper it's just "being initial is invariant under isomorphism in the category of $I$-algebras" but formalizing that literally sounds painful (unless that machinery is already in the library somewhere).

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
fredrik-bakke: Okay then let's leave it for another time 👍

src/Algebra/Lattice/Instances/Free/Inductive.lagda.tree
fredrik-bakke: I had to have a closer look at the code to figure out what I meant. See [b5dc783](https://github.com/samtoth/agda-synthetic-categories/pull/221/commits/b5dc78393735de047c17d6dabb0ed472ce6433c5). (It turns out the argument had to be shifted somewhere else, but at least it is cleaner now. Ultimately we probably want to have this sort of argument at the level of wild categories, but let's not get derailed right now)

ecavallo: I see, that looks nicer. Thanks!

src/Synthetic/Categories/NecklaceLifting.lagda.tree
fredrik-bakke: It sounds like you want to use some instance argument trickery. I'll resolve this issue for now.

PR #226, opened 2026-06-29
An (n+1)-simplex in a lattice is an n-simplex in a coslice lattice

ecavallo:
I want to use this lemma as part of an inductive proof that the n-simplices are spectra.

Please advise me on where it should go. I would have put it in [Algebra.Lattice.Instances.Free.Inductive](https://samtoth.github.io/agda-synthetic-categories/Algebra.Lattice.Instances.Free.Inductive/index.xml), but that module has the lattice as a global parameter and this code needs to talk about simplices in multiple lattices.


src/Algebra/Lattice/Instances/Free/SimplexCoslice.lagda.tree
```diff
@@ -0,0 +1,156 @@
...
+  Δ-coslice
+    : ∀ {n}
+      (xs : Ind.Δ^_ _ M (suc n))
+      → Σ[ y ∶ B ] Ind.Δ^_ _ (Lattice-coslice M y) n
+  Δ-coslice (xs Ind.:≥: y) = (y , Δ-coslice' xs y)
+
+  Δ-uncoslice'
+    : ∀ {n} (y : B)
+      → Ind.Δ^_ _ (Lattice-coslice M y) n
+      → Ind.Δ^_ _ M n
```
fredrik-bakke, line 74:
```suggestion
      → Ind.Δ^ _ (Lattice-coslice M y) n
      → Ind.Δ^ _ M n
```

ecavallo: That won't work: it gets parsed as `Ind.Δ^ (_ M n)` and raises a type error.

These mixfix names don't work well when they're imported from a module with parameters. Note that I could write `(Ind.Δ^ _) M n`, but then I'm superscripting Δ with the carrier of the lattice instead of `n`.

fredrik-bakke: Oh I see, thanks

src/Algebra/Lattice/Instances/Free/SimplexCoslice.lagda.tree
```diff
@@ -0,0 +1,156 @@
...
+      → Ind.last-Δ _ _ (Δ-coslice' xs y ⦃ le ⦄) .fst
+        ＝ Ind.last-Δ _ _ xs
+
+  Δ-coslice' Ind.· y = Ind.·
+  Δ-coslice' ((xs Ind.:≥: x) ⦃ ≤x ⦄) y ⦃ ≤y ⦄ =
+    (Δ-coslice' xs y ⦃ M.≤-trans ≤y ≤x ⦄ Ind.:≥: (x , ≤y))
+      ⦃ subtype-path
+          (M.≤-trans
+            ≤x
+            (M.≤←＝ (sym (Δ-coslice-last xs y ⦃ _ ⦄)))) ⦄
```
samtoth:
```suggestion
  Δ-coslice' ((xs Ind.:≥: x) ⦃ ≤x ⦄) y ⦃ ≤y ⦄ 
    = (Δ-coslice' xs y ⦃ M.≤-trans ≤y ≤x ⦄ Ind.:≥: (x , ≤y))
        ⦃ subtype-path
            (M.≤-trans
              ≤x
              (M.≤←＝ (sym (Δ-coslice-last xs y ⦃ _ ⦄)))) ⦄
```

src/Algebra/Lattice/Instances/Free/SimplexCoslice.lagda.tree
```diff
@@ -0,0 +1,156 @@
...
+\import{stt-macros}
+
+
+\agda{
+open import Foundations.Prelude
+open import Axioms.UF
+
+open import Algebra.Lattice
+open import Algebra.Lattice.Instances.Coslice
+import Algebra.Lattice.Instances.Free.Inductive as Ind
```
samtoth: Is this imported qualified because of name clashes?

src/Algebra/Lattice/Instances/Free/SimplexCoslice.lagda.tree
samtoth:
```suggestion
  Δ-coslice' ((xs Ind.:≥: x) ⦃ ≤x ⦄) y ⦃ ≤y ⦄ 
    = (Δ-coslice' xs y ⦃ M.≤-trans ≤y ≤x ⦄ Ind.:≥: (x , ≤y))
        ⦃ subtype-path
            (M.≤-trans
              ≤x
              (M.≤←＝ (sym (Δ-coslice-last xs y ⦃ _ ⦄)))) ⦄
```

samtoth (approved):
Firstly, really sorry it took so long to get round to reviewing this. 

Reading through this code makes me think that using the name $\Delta^n$ for chains wasn't the best decision and using a non infix name like `Chain L n` would probably make all of this code a bit easier to read. I suggest we merge this without any renaming for now but I will add an issue to think about renaming these things in the future.

src/Algebra/Lattice/Instances/Free/SimplexCoslice.lagda.tree
ecavallo: It seems there isn't any name clash. I don't remember the history now, but maybe it's an artifact of a previous version where there was a clash. I can unqualify it.

PR #229, opened 2026-06-30
An algebra homomorphism from the (n+1)-simplex is a homomorphism from the n-simplex into a slice

ecavallo:
Also:
- a couple of facts about dependent identity types
- edited some of my existing paragraphs to consistently refer to the base lattice as #{L}

I made some definitions opaque to speed up type-checking, but I am pretty clueless about performance issues, so maybe someone else can find a more elegant solution.

### DIFF
--- a/src/Algebra/Lattice.lagda.tree
+++ b/src/Algebra/Lattice.lagda.tree
@@ -415,7 +415,7 @@ LatticeAlgHom-∘ (g , b) (f , a) .snd =
 \subtree[stt-00TB]{
 \taxon{Module}
 \title{Identifying Algebra Homomorphisms}
-
+\contributor{ecavallo}
 
 \p{Being an algebra homomorphism is a property of a map, and hence
   there is an identity system identifying identities of algebra homomorphisms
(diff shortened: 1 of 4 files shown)


fredrik-bakke:
> I made some definitions opaque to speed up type-checking, but I am pretty clueless about performance issues, so maybe someone else can find a more elegant solution.

Are you experiencing performance issues?

ecavallo:
If I don't make `Δ-slice-hom`, `Δ-unslice-hom`, `Δ-slice-unslice-hom`, and `Δ-unslice-slice-hom` opaque, then my Agda starts chugging checking `Δ-slice-hom-is-equiv`.

fredrik-bakke:
Setting the option `--lossy-unification` will probably fix it. Otherwise keeping certain definitions opaque is also fine imo

ecavallo:
`--lossy-unification` does work (with some tweaking), but if it's my choice I prefer `opaque`, which I think is the more local solution.

samtoth:
> `--lossy-unification` does work (with some tweaking), but if it's my choice I prefer `opaque`, which I think is the more local solution.

FWIW, use of opaque is my preference over lossy-unification as well.

I did some experimenting and it seems that just making the `is-lattice-hom` part of both `Δ-slice-hom` and `Δ-unslice-hom` opaque (but not the whole definition) is enough to make it check in a reasonable speed on my computer at least (~1900ms for the whole file). This way the 'important' part of the equivalence still computes at least.

ecavallo:
That sounds better. Would you commit your experiment to the PR?

samtoth:
Actually on further thought, 1,522ms (from opaqueing the equational reasoning blocks), puts it as the second slowest module so probably best to keep as is (which was ~500ms on my PC). I'm still not really sure *why* this is so bad - I wouldn't expect it just from looking at the code.

```
Total                                          50,668ms
Miscellaneous                                     690ms
Core.FibredJoin                                 1,923ms
Algebra.Lattice.Instances.Free.Inductive        1,522ms
Synthetic.Categories.Spaces                     1,510ms
Core.OrthogonalClosure                          1,384ms
Core.CanonicalPushouts                          1,289ms
Core.Joins                                      1,163ms
Synthetic.Categories.Precategories              1,075ms
Core.CanonicalSequentialColimits                1,023ms
Foundations.CanonicalSequentialLimits             963ms
Synthetic.Categories.CoComma                      891ms
Synthetic.Categories.ContravariantHomFamilies     884ms
Synthetic.Categories.Homotopy                     882ms
Core.SpanMaps                                     870ms
...
```

Running with interactive highlighting in emacs (with nothing opaque) seems to imply that it hangs on the `(let module M = Algebra)...` part, which is very confusing to me. I am also not really that knowledgeable on Agda perfomance, so I think just having the opaque as you had it is probably our best bet. (I also found that lossy-unification failed to typecheck as it sometimes does)

PR #230, opened 2026-06-30
Add a remark about code length for associativity lemma, and various tiny fixes

fredrik-bakke:
Adds a remark comparing our formalisation to Rzk sHoTT.


samtoth (approved):
Nice addition!

PR #149, opened 2026-05-19
Descent for pushouts

fredrik-bakke:
Shows that displayed cartesian span diagrams classify families over pushouts.

PR #232, opened 2026-06-30
Fibred version of directed univalence and misc edits

fredrik-bakke:
Adds the fibred version of directed univalence, and makes miscellaneous edits such as adding a negated equality operator, renaming `is-tiny-str` to the more appropriate `√-is-tiny-str`, and renames `CoComma` to `Cocomma`.

Pulled out of #149.


src/Synthetic/Categories/Spaces.lagda.tree
```diff
@@ -605,5 +605,40 @@ module _
...
+#{\mathcal{S}^{Δ¹} \to \Sigma_{(A, B : \mathcal{S})}(A \to B)} is an equivalence,
+we immediately get that the canonical fibred map
+#{\Hom_\mathcal{S}(A,B) \to (A \to B)} is an equivalence as well.
+}
+
+\agda{
+  dua→
+    : ∀ {@♭ 𝓤} (A B : Space 𝓤)
+      → Hom (Space 𝓤) A B → (Space.Carrier A → Space.Carrier B)
+  dua→ ._ ._ (mk-hom h refl refl) = tr-𝓢 h
```
samtoth: I wonder if it would be easy to rewrite this so it computes on any morphism? I guess by composing with a transport either side instead. The computation would then still be the same in the case the hom0 and hom1 are refl

fredrik-bakke: Yeah, I was thinking about that too, but do we want it to unfold always?

samtoth: Yeah that's a good question... 🤔
Not really sure what the answer is at the moment. Even if we don't want it to compute, it might be more principled to write the one with better computational behaviour and then make it opaque

fredrik-bakke: well, with your suggested change we would have to start applying the unit law of identification concatenation when using dua then

samtoth: I'm not sure I follow, wouldn't something like the following work:

`dua→ ._ ._ (mk-hom h hom0 hom1) = tr Carrier hom1 ∘ tr-𝓢 h ∘ tr Carrier hom0`

which, if `hom0` and `hom1` are refl, is just equal to `tr-𝓢 h`?

fredrik-bakke: Oh right, never mind me, I had just misassesed the precise definition

samtoth: Actually, better yet: `tr-𝓢` uses `tr-cov`. We should define `tr-𝓢-hom` using `tr-cov-hom`

src/Synthetic/Categories/Cocomma.lagda.tree
```diff
@@ -271,30 +277,31 @@ This is sometimes also called the \em{blunt join}.}
...
+               , (Square-flatten-hor refl) , Square-flatten-hor refl)) where
+    open directed-cocone cc
     open Homᵈ (filler _)
 
 _▻ : ∀ {𝓤} → Type 𝓤 → Type (𝓘 ⊔ 𝓤)
 A ▻ = A ◇ 𝟙
 
 ▻-span : ∀ {𝓤} (A : Type 𝓤) → Span (𝓘 ⊔ 𝓤) (𝓘 ⊔ 𝓤) 𝓤
-▻-span A = ◇-span A 𝟙
+▻-span A = ◇-span {𝓥 = lzero} A 𝟙
```
samtoth: Do you know why this is now necessary? I'm guessing `--lossy-unification`. How much does turning it off affect this file?

fredrik-bakke: oh, it works fine without. I'll revert. Sorry for the messy code

samtoth (approved):

fredrik-bakke:
Merging this; there are no trees to rename.

PR #233, opened 2026-06-30
Add GitHub tags and OrcIDs for people

fredrik-bakke:


fredrik-bakke:
Here's how it displays
image

PR #234, opened 2026-06-30
Directedness

fredrik-bakke:
Factors out _directedness of the interval_, i.e., that every endomap on it is monotone, and collects some of the surrounding results. As a consequence we no longer need distributivity of the interval again and the world is back in order. We also introduce the condition that $\Delta^2$ classifies monotones on $\Delta^1$, which is the precise missing piece between directedness and synthetic quasicoherence.


src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,165 @@
+\title{Directedness of the interval}
+\author{fredrikbakke}
+\taxon{Module}
+\import{stt-macros}
+
+
+\subtree[fre-00US]{
+\taxon{Idea}
```
samtoth, line 8: I think it would be good to add a justification of the terminology here

fredrik-bakke: done

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,165 @@
...
+We consider directedness of the interval, which is the property that every
+endofunction on the interval is monotone. In other words, every construction respects
+the ordering on #{Δ¹}. This condition [generalises](stt-00TV)
+[synthetic quasicoherence](Synthetic.Quasicoherence), and is at the same time enough
+to derive many important constructions.
+}
+
+\p{
+Directedness boils down to assuming a canonical equivalence
+#{\Hom(\Delta^1,\Delta^1) \simeq (\Delta^1 \to \Delta^1)}, while synthetic
```
samtoth: Is `\Hom` here meant to be a morphism of distributive lattices? Maybe there's a way of writing this more clearly?

fredrik-bakke: Clarified it to `\Hom_{\mathrm{Poset}}(\Delta^1,\Delta^1)`!

src/Synthetic/Directedness.lagda.tree
fredrik-bakke: I also added a disclaimer

```tree
\p{
\strong{Note!} This terminology is not established in the literature, and may be
subject to change in the future.
}
```

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+  → is-monotone (Lattice.≤-poset I) (Lattice.≤-poset I) (λ x → (s ∧ x) ∨ t)
+lin-is-monotone s t a≤b =
+  ∨-UP
+    ( ∨-≤-introl (∧-UP (∧-≤-eliml ≤-refl , ∧-≤-elimr a≤b))
+    , ∨-≤-intror ≤-refl)
+}
+}
+
+\subtree[fre-9F01]{
+\title{#{Δ²} classifies monotone maps on #{Δ¹}}
```
samtoth: As written this sounds like a statement/theorem,rather than a property. Maybe something like 
```suggestion
\title{#{Δ²} classifying monotone maps on #{Δ¹}}
```
?

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+\em{#{Δ²} classifies monotone maps on #{Δ¹}}.
+}
+
+\agda{
+Δ²←Δ¹-monotone
+  : PosetHom (Lattice.≤-poset I) (Lattice.≤-poset I) → Δ² I
+Δ²←Δ¹-monotone (f , f-monotone) = (f i1 , f i0) ⦃ f-monotone 0-init ⦄
+
+Δ²-classifies-Δ¹-monotones : Type 𝓘
+Δ²-classifies-Δ¹-monotones = is-equiv Δ²←Δ¹-monotone
```
samtoth: This could be its own tree

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+
+
+\subtree[fre-00US]{
+\taxon{Idea}
+\author{fredrikbakke}
+
+\p{
+We consider directedness of the interval, which is the property that every
+endofunction on the interval is monotone. In other words, every construction respects
+the ordering on #{Δ¹}. This condition [generalises](stt-00TV)
```
samtoth: Is the link at generalises pointing to the correct place? Currently it points to the definition of directedness

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+}
+}
+
+\subtree[fre-EB55]{
+\taxon{Lemma}
+\author{nimarasekh}
+\date{2026-06-11}
+
+\agda{
+lin-is-monotone
```
samtoth: Maybe this could have a clearer name. `polynomials-are-monotone` or something? (Since all lattice polynomials are linear)

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+\contributor{nimarasekh}
+\contributor{reidbarton}
+
+\p{
+We say the interval #{Δ¹} is \em{directed} if every endofunction on #{Δ¹} is
+monotone. In other words, every construction on #{Δ¹} respects its ordering.
+}
+
+\agda{
+Δ¹-is-directed : Type 𝓘
```
samtoth, line 83: I wonder if this should be `is-directed` and take a lattice as an argument. I guess this would be a bit of work to change since you'd have to remove the lattice parameter from the module.

Even if the module parameter remains unchanged, I think the name without \Delta^1 might be better.

src/Synthetic/Directedness.lagda.tree
samtoth: I mean the lemma about retracts should be separate from the definition

src/Synthetic/Directedness.lagda.tree
```diff
@@ -0,0 +1,170 @@
...
+open import Foundations.Prelude
+open import Core.Coslices
+open import Algebra.Lattice.Instances.Pointwise
+open import Modalities.Instances.Truncation
+
+open import Ergonomics.Extensionality
+}
+}
+
+\subtree[stt-00TV]{
```
samtoth: This tree address is duplicated with one in Synthetic.Quasicoherence

samtoth (commented):
I am still not 100% clear on why the property is called directedness. Is this directed in the same sense of the word as "directed type theory"?

src/Synthetic/Directedness.lagda.tree
fredrik-bakke: yeah that link placement doesn't make sense to me either. Removed it

fredrik-bakke:
> I am still not 100% clear on why the property is called directedness. Is this directed in the same sense of the word as "directed type theory"?

Yeah... idk. It just needs a name and this is what Reid was calling it. Also, it clearly implies some "direction", right? In the sense that we have that endomaps on the interval cannot break its direction

src/Synthetic/Directedness.lagda.tree
fredrik-bakke: and call Delta^2 something else in this module too?

fredrik-bakke:
Directedness does for instance rule out the Kan involution, since that reverses the order

src/Synthetic/Directedness.lagda.tree
samtoth: Okay, upon later reflection I don't think my original suggestion makes sense in the way we currently have the library organised. E.g. It would also make sense to rename things in quasicoherence so they aren't specifically about simplicial type theory. I think this could be worth doing in the future but not for this PR so will resolve this for now

src/Synthetic/Quasicoherence.lagda.tree
```diff
@@ -297,45 +302,58 @@ module _ (H : phoas-principle) where
 \p{
  We show that Phoa's principle implies that every endomorphism of #{Δ¹} is monotone.
  This is a consequence of the fact that the monotone maps from #{Δ¹} to #{Δ¹}
```
samtoth: I guess this should reference directedness in the title (or at least the text) now

samtoth (approved):
Looks good :)

PR #241, opened 2026-07-02
Improvements to typing wrappers

fredrik-bakke:
Adds guidelines for typing wrappers, including fixity, inlining, and display. Implements these guidelines.

This removes some of the awkward occurences where the Agda solver yields operators of the form `ε♭: ...` instead of `ε♭`.


fredrik-bakke:
I also added display pragmas and experimental guidelines for these now.

CONTRIBUTING.md
```diff
@@ -34,6 +34,96 @@ with some extensions.
...
+
+  ```agda
+  infix 25 id:_
+  ```
+
+  This lets us write `id: fibre f y` instead of `id: (fibre f y)`. Examples include
+  `id:_`, `~refl:_`, `fst:_`, and `ε♭:_`.
+
+  Note that while all of these return functions themselves, we rarely if ever have to
+  provide the typing information when we evaluate them. Therefore it is safe to
```
samtoth: I'm a bit unsure what this sentence is saying. What does evaluate mean in this context?

CONTRIBUTING.md
```diff
@@ -34,6 +34,96 @@ with some extensions.
...
+  define these as unary prefix operators.
+
+  We do not make wrappers prefix operators when they are normally used with further
+  explicit arguments. For example, `♭-ind:` is not made a unary prefix operator,
+  since the common use case would otherwise have to be parenthesized like
+  `(♭-ind: B) f`.
+
+- **Inlining.** Typed wrappers should be invisible in computations, so we mark them
+  `INLINE`:
```
samtoth, line 71: I'm still not 100% clear on the purpose of inlining. Reading from the Agda docs:

> A function definition marked with an INLINE pragma is inlined during compilation. If it is a simple function definition that does no pattern matching, it is also inlined in function bodies at type-checking time.
>
> When the [--auto-inline](https://agda.readthedocs.io/en/latest/tools/command-line-options.html#cmdoption-auto-inline) [command-line option](https://agda.readthedocs.io/en/latest/tools/command-line-options.html#command-line-options) is enabled, function definitions are automatically marked INLINE if they satisfy the following criteria:
>
> - No pattern matching.
>
> - Uses each argument at most once.
>
> - Does not use all its arguments.
>
> Automatic inlining can be prevented using the NOINLINE pragma.

We don't do any kind of compilation in this library so that part is irrelevant, I guess the second part might speed up type checking times? I believe we now have `--auto-inline` enabled. Do these typed wrappers satisfy the above criteria anyway (it seems maybe only the first two are satisfied)? If not, are you able to observe any improvement in type checking time with the INLINE pragmas enabled?

CONTRIBUTING.md
fredrik-bakke: evaluating the returned function. E.g., if we evaluate `id: X` at `a`, then the typing argument `X` becomes redundant and we should've just written `id a`.

samtoth: Ah okay I see – thanks

CONTRIBUTING.md
fredrik-bakke: > I guess the second part might speed up type checking times? 

The point here is to try to make the syntax invisible to the user for type-checking purposes, not to speed anything up.

CONTRIBUTING.md
fredrik-bakke: I added the example to the text.

CONTRIBUTING.md
samtoth: Hmm okay, I am under the impression that INLINE has no effect on the user's interaction with Agda

samtoth: For example – in the Agda-zulip, all mentions of inlining I have found so far are about appeasing the termination checker

fredrik-bakke: :( I'll reevaluate

fredrik-bakke: maybe `syntax` is the correct tool to use here instead

samtoth: Do the display pragmas on their own not do the job? It seems like they are the intended tool for improving goal display

fredrik-bakke: No, as pointed to there's an issue with display pragmas for functions returning functions.

fredrik-bakke: Oh. Maybe I've misunderstood the DISPLAY pragma

fredrik-bakke:
don't have the interest to work on this right now, so I've reconverted it to a draft.

PR #244, opened 2026-08-18
Segal types are left and right Segal

fredrik-bakke:
We show that the Segal condition is enough to prove that the hom family is appropriately covariant and contravariant, and hence that the type is resp. left and right Segal. As consequences, we get the Yoneda lemma for Segal types and the strictification construction.

Also removes distributivity hypothesis from Yoneda and Strictification again.


src/Synthetic/Categories/ContravariantFamilies.lagda.tree
```diff
@@ -448,6 +451,151 @@ module _ {𝓤 𝓥} {A : Type 𝓤} (P : A → Type 𝓥) (Pc : is-contravarian
...
+\proof{
+\p{
+The forwards direction is obvious. For the backwards direction we show that the
+family is a right family by constructing an explicit inverse. I.e., we must
+construct a lift of the diagram
+\quiver{
+\begin{tikzcd}
+\{1\} && \Sigma_{(a:A)} B(a) \\
+\\
+Δ¹ && A
```
samtoth:
```suggestion
\Delta^1 && A
```

src/Synthetic/Categories/CovariantFamilies.lagda.tree
```diff
@@ -473,6 +477,148 @@ opaque
...
+\proof{
+\p{
+The forwards direction is obvious. For the backwards direction we show that the
+family is a left family by constructing an explicit inverse. I.e., we must construct
+a lift of the diagram
+\quiver{
+\begin{tikzcd}
+\{0\} && \Sigma_{(a:A)} B(a) \\
+\\
+Δ¹ && A
```
samtoth:
```suggestion
\Delta^1 && A
```

samtoth (commented):
(Not a proper review, just fixes build step)

src/Synthetic/Categories/CovariantFamilies.lagda.tree
```diff
@@ -473,6 +477,148 @@ opaque
...
+For each #{i : Δ¹}, let #{0\leq i : \Hom_{Δ¹}(0,i)} be the canonical morphism.
+Applying #{f} we get a morphism #{f(0\leq i) : \Hom_A(f(0),f(i))} along which we
+can transport #{b}, and get
+##{f'(i) \coloneqq \left(f(i), f(0\leq i)_*(b)\right).}
+This lies over #{f}, and its value at #{0} is #{(f(0),b)} by unitality.
+}
+
+\p{
+This construction is an inverse to restriction at #{0}. One inverse law follows
+immediately from unitality. For the other, the contraction of #{Δ¹_{/i}} from
```
samtoth, line 545: Maybe it's worth highlighting (at least in the prose) that a unital transport operation is the same as a _section_ to the corresponding gap map. It might also be worth extracting this as a lemma in the code, but feel free to leave this to future work. (I want to formalise the analogous theorem for cocartesian fibrations where this would be useful so can do it then for example)

fredrik-bakke: I'm happy to defer it to when you want to work on this, but if you prefer it if I add it here I can do that too. The proof as written is how I interpreted Rob's intention of what the proof was.

src/Synthetic/Categories/Yoneda.lagda.tree
```diff
@@ -168,103 +168,31 @@ any family #{P : A \to \UU} into a family #{{\pi_1}^*P : A_{a/} \to \UU}.}
...
-  Yoneda-coe-left-segal-is-equiv
-    : (yon : P a → (x : A) → Hom A a x → P x)
-    → ((p0 : P a) → yon p0 a (idH a) ＝ p0)
-    → is-equiv yon
-  Yoneda-coe-left-segal-is-equiv
-    = YonedaJ-left-segal-is-equiv 𝐴ₗ _ cov'
-}
-}
-
-\subtree[stt-00FT]{
```
samtoth, line 200: Have references to this tree (if there are any) been renamed to refer to STT-00LK instead?

fredrik-bakke: good catch!

trees/stt/stt-00VI.tree
samtoth, line 1: Take or leave the following but there could be some more 'colour' to be added here.

A) that it was kinda surprising that these things turned out to be equivalent (at least it wasn't obvious to me that it would be the case)

B) maybe you could briefly mention what motivated the definition of left/right Segal, e.g the original proof in RS used the full power of simpliciality, which was then reduced to just the left Segal condition and then reduced to just Segal.

fredrik-bakke: I agree but don't have time to think up something rn 😭

src/Synthetic/Categories/ContravariantHomFamily.lagda.tree
```diff
@@ -200,10 +201,37 @@ hom-fam-is-contrav↔is-right-segal {𝓤} {A}
...
+  = is-contravariant←contrav-unital-transport
+      (λ _ → Hom-is-Δ¹-discrete←is-segal 𝐴)
+      (precomp-hom 𝐴 z , (λ _ → ∘=←Hom2 𝐴 ∘ refl~▵))
+
+is-right-segal←is-segal : ∀ {𝓤} {A : Type 𝓤} → is-segal A → is-right-segal A
+is-right-segal←is-segal 𝐴
+  = hom-fam-is-contrav↔is-right-segal .fst (hom-fam-contrav 𝐴)
+
+is-segal↔is-right-segal : ∀ {𝓤} {A : Type 𝓤} → is-segal A ↔ is-right-segal A
+is-segal↔is-right-segal = (is-right-segal←is-segal , is-segal←is-right-segal)
```
samtoth, line 235: Could be separated as a corollary tree

src/Synthetic/Categories/CovariantHomFamily.lagda.tree
```diff
@@ -244,39 +249,37 @@ coslice-is-left↔is-segal←triangulated Atri
 \taxon{Theorem}
 
 \p{
-Given a [left Segal](stt-00DA) type #{A}, the covariant action of
+Given a [Segal](stt-00AI) type #{A}, the covariant action of
 the family #{(a : A) \mapsto \Hom_A(a,{-})} is given by postcomposition.
 }
```
samtoth, line 254: I wonder if this proof can be simplified now. E.g. it might hold by definition. Not needed for this PR but will make an issue to investigate this at some point

trees/stt/stt-00DS.tree
```diff
@@ -40,6 +40,10 @@ setting (where we don't assume #{I} to be totally ordered) this
...
 local wrt. the inclusion #{\Lambda^2_1 \to 1 * \Delta^1} (called
-being [(left) strong Segal](stt-00DA)) iff #{\Hom(x,-)} is covariant.
- We can then separately show that [simplicial types are Strong Segal
-iff they are Segal](stt-00DG).}
+being [left Segal](stt-00DA)) iff #{\Hom(x,-)} is covariant.
+We can then separately show that simplicial types are Strong Segal
+iff they are Segal.}
+
+\remark{
+The final point above has been made redundant following [[fre-1010]].
```
samtoth:
```suggestion
The final point above has been made redundant following the result that [[fre-1010]].
```

src/Synthetic/Categories/LeftSegal.lagda.tree
```diff
@@ -227,122 +231,3 @@ opaque
               (Pushout-assoc→-is-equiv Δ²-diagonal Δ²-diagonal (_, i0) !))))
 }
 }
-
```
samtoth, line 230: I wonder if it's worth keeping any of the intermediate lemmas here.

samtoth (approved):

src/Synthetic/Categories/ContravariantHomFamily.lagda.tree
fredrik-bakke: I agree, but don't want to create and rename identifiers again rn. Leaving this for the future.

src/Synthetic/Categories/LeftSegal.lagda.tree
fredrik-bakke: It's in the commit log, so we can restore it later and consider

src/Synthetic/Categories/ContravariantHomFamily.lagda.tree
fredrik-bakke: Fixed in #234

PR #249, opened 2026-08-26
Semilattices are sets

fredrik-bakke:
Formalises that the algebraic structure of a semilattice implies that the carrier type is a set.
The result is used to slightly simplify the development of lattices.

This result is due to David Wärn. Parts of the accompanying formalisation follows Tom de Jong's [streamlined formalisation](https://martinescardo.github.io/TypeTopology/AlgebraicStructuresForcingSethood.Semilattices-streamlined.html) in the TypeTopology library, and is hence licensed under GPL 3.0.

PR #252, opened 2026-08-28
The Eckmann–Hilton argument

fredrik-bakke:
Formalises the algebraic Eckmann–HIlton argument, and shows that concatenation on double loops is commutative.

Builds on #251 and #249.


src/Foundations/EckmannHilton.lagda.tree
```diff
@@ -0,0 +1,224 @@
...
+    → (x y : X)
+    → x ◽ y ＝ y ◽ x
+comm←interchange _◾_ _◽_ e l◾ r◾ l◽ r◽ ◾-◽-interchange x y
+  = equal-ops←outer-interchange
+      (_◾_) (_◽_) e l◾ r◾ l◽ r◽
+      (λ a d → ◾-◽-interchange a e e d)
+      (x)
+      (y)
+  ∙ sym
+      ( opposite-ops←inner-interchange
```
samtoth:
```suggestion
      (opposite-ops←inner-interchange
```

src/Foundations/EckmannHilton.lagda.tree
```diff
@@ -0,0 +1,224 @@
+\date{2026-08-27}
+\title{The Eckmann–Hilton argument}
+\author{fredrikbakke}
+\taxon{Module}
+\import{stt-macros}
+
+\def\hconcat{\mathbin{◾}}
+\def\vconcat{\mathbin{◽}}
+
```
samtoth, line 9: If you'd be willing, a short `\idea` section could be nice here

src/Foundations/EckmannHilton.lagda.tree
```diff
@@ -0,0 +1,224 @@
+\date{2026-08-27}
+\title{The Eckmann–Hilton argument}
+\author{fredrikbakke}
+\taxon{Module}
+\import{stt-macros}
+
+\def\hconcat{\mathbin{◾}}
+\def\vconcat{\mathbin{◽}}
```
samtoth, line 8: Just thinking, these won't be available in other modules which import this one via Agda, unless they also `\import{Foundations.EckmannHilton}`. Maybe you don't need to use these macros in other files, but if you do it might be better to put them in `stt-macros`

samtoth (approved):
Great :)

src/Foundations/EckmannHilton.lagda.tree
fredrik-bakke: the intention was to only use this notation locally here

samtoth:
I went to try and merge this, but it is a little complicated since some other PR defines some of the same things but in different files. Would you mind doing this merge?

fredrik-bakke:
Yep, I'll do it shortly!

PR #253, opened 2026-08-29
Syllepsis

fredrik-bakke:
The Eckmann–Hilton symmetry $\mathsf{EH}$ is its own inverse on 3-loops. I.e., given two 3-loops $p$ and $q$, then $\mathsf{EH}(p,q) = \mathsf{EH}(q,p)^{-1}$.

Follows Sojakova–Kavvos https://dl.acm.org/doi/epdf/10.1145/3531130.3533347.

Builds on #252, #251, and #249.


fredrik-bakke:
I anticipate that the proof of syllepsis is considerably simpler/shorter for binary operations that have strictly unital whiskering, such as the Yoneda system. This is because most of the reasoning is to handle degenerate squares,  which boils down to dealing with the unit laws of whiskering!

fredrik-bakke:
Here's an excerpt of the rendering
image

src/Foundations/EckmannHilton.lagda.tree
```diff
@@ -6,6 +6,15 @@
 
 \def\hconcat{\mathbin{◾}}
 \def\vconcat{\mathbin{◽}}
+\def\hpaste{\mathbin{◫}}
+\def\vpaste{\mathbin{⊟}}
```
samtoth: I wonder if these two could go in stt-macros, since they are defined 'globally' and in a different file?

fredrik-bakke: yep good idea

fredrik-bakke: oh, actually, I need to find their latex commands, because these macros will break tikzcd

samtoth: Yeah... If you need to define a LaTeX macro that can be done in the 'quiver prelude' in `trees/macros/basic-macros`. But it might be tricky because you will need a macro that works both with Katex and whatever latex backend forester uses. (I wonder if it can be configured to use a different backend that supports unicode....)

fredrik-bakke: yea apparantly I cant get it done for horizontal pasting without a custom glyph for katex. I've just added a warning for it for now not to use with tikzcd

fredrik-bakke:
We should get a custom macro for path concatenation. the cdot is way too small

PR #254, opened 2026-08-30
Move truncated types to foundations

### LINKED ISSUE #164: Factor out truncatedness from `Modalities`
Currently truncatedness depends on `Axioms.UF` and the whole ordeal, which means that we can't state things like that the natural numbers form a set without importing all of univalent foundations. Instead we should have a module `Foundations.Truncation` that does not import any of those things but still includes the basic definitions.

fredrik-bakke:
- Factors out new modules `Foundations.Truncated` and `Foundations.Sets`.
- Also adds some useful results about sets, such as Hedberg's theorem.
- Also adds `Foundations.InjectiveMaps`

Resolves #164.


src/Foundations/InjectiveMaps.lagda.tree
```diff
@@ -0,0 +1,91 @@
...
+}
+
+\subtree[fre-9001]{
+\title{Injective maps}
+\taxon{Definition}
+
+\p{
+A map #{f : A \to B} is \em{injective} if from #{f(x) = f(y)} it follows that
+#{x = y} for any #{x\ y : A}.
+}
```
samtoth, line 29: Remark that "*is*-injective" is a slight abuse of terminology. (but an acceptable one since it is a proposition in common cases)

src/Foundations/Truncated.lagda.tree
```diff
@@ -0,0 +1,425 @@
...
+      → is-truncated n A
+      → is-truncated n (ULift 𝓥 A)
+  ULift-is-truncated {𝓥 = 𝓥} tr
+    = is-truncated←equiv (lift≃ {𝓥 = 𝓥}) tr
+
+  is-truncated←is-singleton
+    : ∀ {𝓤} {n : 𝕋} {A : Type 𝓤}
+      → is-singleton A
+      → is-truncated n A
+  is-truncated←is-singleton sing = is-truncated←is-single sing _
```
samtoth: The difference here is in whether `n` is implicit. Maybe `is-truncated←is-single` should take an implicit truncation level, and then this def can be removed

src/Foundations/Truncated.lagda.tree
```diff
@@ -0,0 +1,425 @@
...
+    → is-truncated n (A × B)
+×-is-truncated Atr Btr = Σ-is-truncated Atr (λ _ → Btr)
+
+Pullback-is-truncated'
+  : ∀ {𝓦 𝓧 𝓨} {n : 𝕋}
+      {X : Type 𝓦} {Y : Type 𝓧} {Q : Type 𝓨}
+      (g : X → Q) (h : Y → Q)
+    → is-truncated n X
+    → is-truncated n Y
+    → (∀ {x y : Q} → is-truncated n (x ＝ y))
```
samtoth: Is there a good reason not to use `is-truncated (suc n) (x ＝ y)` here? I guess implicit-ness, but I don't know if its better to just have explicit arguments instead

src/Foundations/Truncated.lagda.tree
```diff
@@ -0,0 +1,425 @@
...
+    → is-truncated n (Pullback g h)
+Pullback-is-truncated g h Xtr Ytr Qtr
+  = Pullback-is-truncated' g h Xtr Ytr (＝-is-truncated Qtr)
+
+fibre-is-truncated'
+  : ∀ {𝓦 𝓧} {n : 𝕋}
+      {X : Type 𝓦} {Y : Type 𝓧}
+      (g : X → Y) (y : Y)
+    → is-truncated n X
+    → (∀ {x y : Y} → is-truncated n (x ＝ y))
```
samtoth: Same comment (as pullback-is-truncated) here

samtoth (approved):
👍 👍

src/Foundations/Truncated.lagda.tree
fredrik-bakke: agree

src/Foundations/InjectiveMaps.lagda.tree
fredrik-bakke: I can rename it to `injective`, since, commonly, in homotopy type theory, it will be a non-proposition in interesting ways

samtoth: Hmm, I think I prefer `is-injective` still (with an accompanying remark). We do similar abuses in other places, e.g. `is-small`. There are some better examples I’m sure, but I can't think of any right now.

fredrik-bakke: okies. `is-small` is a property though

samtoth: Only assuming univalence was my point... but It's not exactly the same scenario as this

PR #260, opened 2026-09-03
Generalise to types with initial objects separate

fredrik-bakke:
Just noticed we don't ever need induction on $n$ for our cubes separate argument. ~~I suppose this edit is sort of redundant if it is readily possible to use only that types with initial objects separate...~~

EDIT: Following a further suggestion by Sam, the contribution has been generalised further to the assumption that types with initial objects separate.


samtoth:
I think I will merge this after merging HornGlueing into main – if that's okay

fredrik-bakke:
Totally!

fredrik-bakke:
@samtoth I generalised this contribution to types with initial objects separate now

src/Synthetic/Categories/EquivCovariantFamilies.lagda.tree
```diff
@@ -48,16 +45,113 @@ open import Core.ArrowEquiv
...
+open import Synthetic.Categories.Precategories Δ¹ I
 open import Synthetic.Categories.CovariantFamilies Δ¹ I
+open import Synthetic.Categories.Initial Δ¹ I
 }
 }
 
 \subtree[stt-00V6]{
-\title{The point inclusion of #{0} into the #{n}-cube is left anodyne}
-\taxon{Lemma}
+\title{The point inclusion of an initial object is left anodyne}
```
samtoth: I think this should go in a different module. I think more or less the same lemma exists in the Yoneda lemma module (except using lifting against covariant families for its definition of left anodyne). Maybe you could merge these two?

fredrik-bakke:
If we generalise further to types with an object whose based hom has a section, then we can get rid of the Δ¹-is-segal assumption too......

src/Synthetic/Categories/EquivCovariantFamilies.lagda.tree
fredrik-bakke: Done. I moved it to `Initial`. I also renamed `EquivCovariantFamilies` to `DetectingEquivalencesCovariantFamilies`

src/Synthetic/Categories/Initial.lagda.tree
```diff
@@ -159,3 +216,119 @@ coslice-init 𝐴 a
...
+\title{The point inclusion of an initial object is left anodyne}
+\taxon{Lemma}
+\author{fredrikbakke}
+\author{samueltoth}
+\date{2026-09-03}
+
+\p{
+The point inclusions of [initial objects](stt-00FO) are left anodyne.
+}
+
```
samtoth: Could you add a sketch of the argument

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
```diff
@@ -43,29 +41,119 @@ open import Core.ArrowRetracts
...
+
+\subtree[bak-1272]{
+\title{Separation by cubes and types with initial objects}
+\taxon{Lemma}
+\author{fredrikbakke}
+\date{2026-09-06}
+
+\p{
+If types with initial objects separate then cubes separate, and if cubes separate
+then finite cubes separate.
```
samtoth: I think these implications are in reverse order

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
```diff
@@ -118,52 +206,84 @@ retract-Δ²-points-∂Δ¹×∂Δ¹ = mk-arr-retract
...
+
+  ♭Δ²-ind
+    : ∀ {𝓤} (P : @♭ Δ² → Prop 𝓤)
+      → (∀ (@♭ x : Fin 3) → P (Δ²-point-incl x) .fst)
+      → (@♭ x : Δ²) → P x .fst
+  ♭Δ²-ind
+    = up-♭-surjection←up-♭-connected-map
+        (up-connected-map←is-connected-map _
+          (Δ²-points-♭-connected ∂Δ¹-♭-connected))
+}
```
samtoth, line 238: Good idea, this is much cleaner.

src/Synthetic/Categories/Initial.lagda.tree
```diff
@@ -159,3 +216,119 @@ coslice-init 𝐴 a
...
+If #{A} is a type with an [initial object](stt-00FO) #{a}, and #{P} a
+[covariant family](stt-00AY) over #{A}, then the evaluation map at #{a} is an
+[equivalence](stt-001U) #{(\Pi_{x}P(x)) \toeq P(a)}.
+}
+
+\proof{
+\p{
+We first give the inverse map. Given a partial section at #{a}, #{x : P(a)},
+we use the covariance of #{P} and the fact #{a} is initial to transport #{x}
+to a full section: #{(x : P(a)) \mapsto ¡_{b}^*(x) : P(b)}. This map is
```
samtoth: I believe I wrote this in the past but I think the prose is wrong, should read:


```suggestion
to a full section: #{(b : A) \mapsto ¡_{b}^*(x) : P(b)}. This map is
```

src/Synthetic/Categories/Yoneda.lagda.tree
samtoth:
```suggestion
initial object by \ref{stt-00LJ}. Then the result follows directly from
\ref{stt-00FS}.
```

src/Synthetic/Categories/Initial.lagda.tree
```diff
@@ -159,3 +216,119 @@ coslice-init 𝐴 a
...
+of the uniqueness of covariant transport.
+}
+}
+
+\p{
+We give a mild generalisation of \citet{Thm 9.7}{RS17}. Namely, we formalise the
+result under the assumption of a family of morphisms
+#{h_{(-)} : \Pi_{(b : A)} \Hom_A(a,b)} such that #{h_a = \id_a}. When the underlying
+type is Segal, this condition is equivalent to initiality of #{a} and recovers
+the cited theorem.
```
samtoth, line 247: Unsure about "recovers the cited theorem". To me this makes it sound like Segalness is assumed in the theorem of Riehl and Shulman.

src/Synthetic/Categories/DetectingEquivalencesCovariantFamilies.lagda.tree
```diff
@@ -43,29 +41,119 @@ open import Core.ArrowRetracts
...
 open import Synthetic.Horns Δ¹ I
 open import Synthetic.Boundaries Δ¹ I
 open import Synthetic.Triangles Δ¹ I
+open import Synthetic.Categories.Precategories Δ¹ I
 open import Synthetic.Categories.CovariantFamilies Δ¹ I
+open import Synthetic.Categories.Initial Δ¹ I
 }
 }
 
-\subtree[stt-00V6]{
```
samtoth, line 54: Could you search for references to this deleted tree, and replace them with a the relevant new tree.

src/Synthetic/Categories/Initial.lagda.tree
```diff
@@ -48,6 +55,56 @@ is a singleton.
...
+
+initial-obj : ∀ {𝓤} → Type 𝓤 → Type (𝓘 ⊔ 𝓤)
+initial-obj A = Σ A is-initial
+
+Type-With-Initial-Obj : (𝓤 : Level) → Type (𝓘 ⊔ lsuc 𝓤)
+Type-With-Initial-Obj 𝓤 = Σ (Type 𝓤) initial-obj
+
+Type-With-Initial-Obj' : (𝓤 : Level) → Type (𝓘 ⊔ lsuc 𝓤)
+Type-With-Initial-Obj' 𝓤
+  = Σ[ A ∶ Type 𝓤 ] Σ[ a ∶ A ] Σ[ h ∶ ((y : A) → Hom A a y) ] (h a ＝ idH a)
```
samtoth, line 70: I was trying to think if there was a concise and correct name with which we could refer to this condition. But it's difficult because more than just a section of the map `A_{x/} -> A`. I think its a `pointed section` but that doesn't make for a descriptive or snappy name. I guess this is fine as it is initial under the right conditions...

src/Synthetic/Categories/Initial.lagda.tree
```diff
@@ -159,3 +216,119 @@ coslice-init 𝐴 a
...
+\date{2026-02-10}
+\taxon{Lemma}
+\author{samueltoth}
+\contributor{fredrikbakke}
+
+\p{
+If #{A} is a type with an [initial object](stt-00FO) #{a}, and #{P} a
+[covariant family](stt-00AY) over #{A}, then the evaluation map at #{a} is an
+[equivalence](stt-001U) #{(\Pi_{x}P(x)) \toeq P(a)}.
+}
```
samtoth, line 231: Would you a mind adding a remark that this condition is the same as saying that `{a} -> A` is left orthogonal to any covariant family `P`. Or maybe add the remark in the following tree and change it's taxon to corollary?

src/Synthetic/Categories/Yoneda.lagda.tree
```diff
@@ -106,8 +42,8 @@ We first prove the dependent Yoneda lemma. Given a [Segal](stt-00AI) type #{A},
 an element #{a : A}, and some covariant family #{P : \Pi_{x} \Hom(a,x) \to \UU}, we
 show that the evaluation map #{(\Pi_{x}\Pi_{f : \Hom(a,x)} P(x,f)) \to P(a,\id)} is
 an equivalence. If we uncurry #{P}, we get a family #{A_{a/} \to \UU}, which has an
```
samtoth, line 44:
```suggestion
an equivalence. If we uncurry #{P}, we get a family #{A_{a/} \to \UU}, the base of which has an
```

samtoth (approved):

src/Synthetic/Categories/Initial.lagda.tree
fredrik-bakke: > I guess this is fine as it is initial under the right conditions...

Yea, I think this is quite fine.

src/Synthetic/Categories/EquivCovariantFamilies.lagda.tree
```diff
@@ -43,29 +41,119 @@ open import Core.ArrowRetracts
...
 open import Synthetic.Horns Δ¹ I
 open import Synthetic.Boundaries Δ¹ I
 open import Synthetic.Triangles Δ¹ I
+open import Synthetic.Categories.Precategories Δ¹ I
 open import Synthetic.Categories.CovariantFamilies Δ¹ I
+open import Synthetic.Categories.Initial Δ¹ I
 }
 }
 
-\subtree[stt-00V6]{
```
fredrik-bakke: It's not deleted, it's moved to `Initial`

src/Synthetic/Categories/Initial.lagda.tree
fredrik-bakke: good catch

src/Synthetic/Categories/Initial.lagda.tree
fredrik-bakke: I've registered it as a corollary. In my mind, the next subtree _is_ the remark.

PR #262, opened 2026-09-03
Phoa's principle for Δ²

fredrik-bakke:
Parts extracted from #77 and #225.

PR #267, opened 2026-09-07
Reorganise quasicoherence

### LINKED ISSUE #245: Fix terminology in Synthetic.Quasicohence
I believe that we were following the convention in Sterling and Ye's [Domains and Classifying Topoi](https://arxiv.org/abs/2505.13096) in the module [[Synthetic.Quasicoherence]](https://samtoth.github.io/agda-synthetic-categories/Synthetic.Quasicoherence)

The terminology has since been updated in this paper (see Jon's writeup on the matter here: https://www.jonmsterling.com/0LVV/) and so I think we should update our terminology in this module to match.

In particular, the suggestion is that what we were previously calling quasicoherent algebras should be called spatial algebras.

This might be of interest to @markrd-williams (as the original author of the quasicoherence module)

samtoth:
Resolves #245


fredrik-bakke:
Please merge #262 before this.

PR #276, opened 2026-09-09
No need to confirm if no changes — `assign-tree-ids`

fredrik-bakke:
Makes it so there is no ask for confirmation if assigning tree ids does nothing. Also, it will no longer run if the working branch is dirty. Also, ignores `autogen` files now. Also, some formatting improvements to the output. Also, uses the predefined `EXT` variable now.


samtoth:
Have you tested this? If yes, then happy to merge.

fredrik-bakke:
> Have you tested this? If yes, then happy to merge.

yup

PR #277, opened 2026-09-09
Types that are null at pointed types are closed under injections

fredrik-bakke:
If `Y` is pointed, then if `B` is `Y`-null and `f : A -> B` is injective, then `A` is also `Y`-null. In particular, Δ¹-discrete types and rezk-complete types are closed under injections.

### DIFF
--- a/src/Modalities/Instances/Nullification.lagda.tree
+++ b/src/Modalities/Instances/Nullification.lagda.tree
@@ -88,14 +88,6 @@ Null-SubU A 𝓦 .Subtype.family-is-subtype = is-null-is-prop
 \taxon{Proposition}
 \author{fredrikbakke}
 
-\p{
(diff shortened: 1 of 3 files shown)


src/Modalities/Instances/Nullification.lagda.tree
```diff
@@ -82,6 +83,30 @@ Null-SubU A 𝓦 .Subtype.family-is-subtype = is-null-is-prop
 }
 
 
+\subtree[bak-0000]{
+\title{Types that are null at pointed types are closed under path-cosplittings}
```
samtoth, line 87:
```suggestion
\title{Types that are null at pointed types are closed under [path-cosplittings](????)}
```

samtoth (approved):

PR #280, opened 2026-09-10
Descent for surjections

fredrik-bakke:
Shows the descent property of surjections.


samtoth (approved):
Aside from me not fully understanding the way this is a descent condition, everything looks good. Very nice documentation as well 👍

PR #297, opened 2026-09-14
Nix build in Makefile

### LINKED ISSUE #296: Nix build should use Makefile
So we don't have to keep two build scripts in sync

jordanrule:
Fixes: https://github.com/samtoth/agda-synthetic-categories/issues/296


fredrik-bakke:
Hey Jordan. Thanks for the submission, it's much appreciated! I've requested a review from Sam as he's the nix expert.

Makefile
```diff
@@ -97,6 +99,16 @@ build-forest: $(EVERYTHING_FILE) prepare-agda-datadir prepare-forest-assets
 	@mkdir -p "$(AGDA_DATADIR)" "$(AGDA_DATADIR)/lib" "$(AUTOGEN_DIR)" "$(HTML_DIR)"
 	@Agda_datadir="./$(AGDA_DATADIR)" agda-forester --forest -o"$(AUTOGEN_DIR)" --fhtml-dir="$(HTML_DIR)" --fhtml-link-root="/agda-synthetic-categories/html/" --fhtml-css-path="../Agda.css" --fforest-root="/agda-synthetic-categories/" --fdisable-backlinks "$(EVERYTHING_FILE)" -j
 
+nix-build:
+	@$(MAKE) --no-print-directory build-forest HTML_DIR=output/html
+	@$(MAKE) --no-print-directory check-forest-no-typecheck HTML_DIR=output/html
```
samtoth, line 104: `check-forest-no-typecheck` has `sync-forest-src` as a dependency – which copies the Agda source trees without rendering links etc. So you will have to just call `forester build`manually here.

Makefile
```diff
@@ -97,6 +99,16 @@ build-forest: $(EVERYTHING_FILE) prepare-agda-datadir prepare-forest-assets
 	@mkdir -p "$(AGDA_DATADIR)" "$(AGDA_DATADIR)/lib" "$(AUTOGEN_DIR)" "$(HTML_DIR)"
 	@Agda_datadir="./$(AGDA_DATADIR)" agda-forester --forest -o"$(AUTOGEN_DIR)" --fhtml-dir="$(HTML_DIR)" --fhtml-link-root="/agda-synthetic-categories/html/" --fhtml-css-path="../Agda.css" --fforest-root="/agda-synthetic-categories/" --fdisable-backlinks "$(EVERYTHING_FILE)" -j
 
+nix-build:
+	@$(MAKE) --no-print-directory build-forest HTML_DIR=output/html
```
samtoth, line 103: Can `build-forest` be made a dependency of this target. I think it would be the more makefile idiomatic way of doing things. The `HTML_DIR` option can be passed from the build script in `default.nix`

samtoth (changes requested):
Thanks for the PR. You'll have to forgive me because I really am no expert in these matters – and the original issue I created did not have much info.

I have a few general thoughts. Currently the makefile target `make build-forest` is poorly named, really what it does is render the Agda source to the autogen directory inside the forest source directory, maybe a better name for it could be `gen-agda-trees` or something (open to suggestions).

Then I think the target you have named `nix-build` could be called `build-forest` or even just `build` as there isn't anything nix specific about it as far as I can tell.

samtoth:
Hi, I've seen you've requested another review. Would you be able to address the comments in my last review first. Happy to discuss or clarify anything.

Issue #287, opened 2026-09-12
Add index pages


fredrik-bakke:
We should have index pages that lists files in certain directories. E.g. a page for references, a page for people, a page for Foundations modules and so on. This makes the website much more browsable.

fredrik-bakke:
My understanding is that forester has some native support for listing files of certain kinds, e.g. references.

samtoth:
> My understanding is that forester has some native support for listing files of certain kinds, e.g. references.

Yes, it has this query system, which I have been meaning to learn about. So this would be a good excuse to do that.

samtoth:
So I have been playing around with this and people and references are very easy to do (they are builtin relations).

It is possible to query for all modules by querying for trees which have the module taxon. The downside is that this creates a very large and very laggy page, since it actually transcludes every module in the library. Maybe it is better to just create links rather that actually transclude. Or maybe we could have one page for each 'namespace', the only way I can see of doing this is annotating each module with a 'tag' (or getting agda-forester) to do the annotation. Let me know if you have any ideas @fredrik-bakke

fredrik-bakke:
Transclusion sounds like a very bad idea indeed. Tagging also sounds a bit impractical, unless we create a script that checks for it or something. I agree one page per namespace is a good idea.

samtoth:
Created an upstream issue: https://github.com/samtoth/agda-forester/issues/17

PR #311, opened 2026-09-16
Add bibliography index

samtoth:
Partial resolution of #287


trees/index.tree
```diff
@@ -18,6 +18,7 @@ a mixture of [literate agda](https://agda.readthedocs.io/en/latest/tools/literat
  \li{Read an [introduction](stt-0001) to the library.}
  \li{Read about [recent changes to the library](changelog).}
  \li{See a [list of all the modules](https://samtoth.github.io/agda-synthetic-categories/html/Everything) in the library.}
+  \li{Browse the [project bibliography](bibliography) and [list of people](people).}
```
fredrik-bakke: IDK if list of people is interesting, but the bibliography sure is nice!

samtoth: Yeah I was thinking the same. I might remove the link here but keep the tree as a little easter egg anyway

samtoth:
I also don't know if there is a nice solution for the module indexes that only uses Forester datalog queries. I might make an issue upstream to have this as an optional feature of Agda-forester

fredrik-bakke (approved):
I haven't built this locally to see, but I'm happy to take your word on it if the index files work.

PR #318, opened 2026-09-18
chore: remove unused imports

fredrik-bakke:
Ran a local script to remove all unused imports in the library. The script is not ready for prime time, but the edits certainly are :)
This should improve typechecking times somewhat while working locally.


fredrik-bakke:
For the record, this change along yields about a 5% performance increase for me locally, when checking the whole library. I don't expect this to have an effect on the benchmarking however, since it runs sequentially.

fredrik-bakke:
We could reconfigure the benchmark to run on two cores if we wish (I think that is all that is made available) per runner.

samtoth:
That's a lot of unusued imports 😵‍💫

PR #320, opened 2026-09-19
Amazing fibrancy lifting

fredrik-bakke:
Formalises the "proof strategy" in closure properties of amazing covariance and uses it to derive the given closure properties.


src/Synthetic/Tiny.lagda.tree
```diff
@@ -747,6 +749,53 @@ former preserves propositions. This is at the moment left for future work.}
...
 }
 
+\subtree[bak-95B3]{
+\title{Amazing fibrancy lifting}
+\taxon{Construction}
+\author{fredrikbakke}
+\date{2026-09-19}
+
+\p{
+Given a crisp type family #{P \colon-flat A \to \UU}, if #{P} is a fibration then
```
samtoth, line 786: I guess this should be, if `#{I} is a pointed tiny object, for some I-relative notion of fibration, ...`. 

At first I didn't understand why `i0` was a parameter at all the use sites, I wonder if this def could be put in a module with an (i : I) as a module parameter to avoid this confusion.

samtoth (approved):
Love this!! So much neater now

src/Synthetic/Tiny.lagda.tree
fredrik-bakke: There's only one lemma here though?

samtoth: What I am trying to say is that when this lemma is used, e.g.:

```agda
opaque
  Π-is-acov
    : ∀ {@♭ 𝓤 𝓥} {@♭ K : Type 𝓤} {P : K → Type 𝓥}
      → ((k : K) → is-acov (P k)) → is-acov (Π K P)
  Π-is-acov {𝓥 = 𝓥} {K} ac
    = lift-fibrancy i0
        (λ (P : K → Space 𝓥) → (k : K) → Space.Carrier (P k))
        (Π-is-covariant (λ k P → Space.Carrier (P k)) (λ k → UFib-is-fib (ev k)))
        (λ k → mk-ufib _ (ac k))
```

we have to write `lift-fibrancy i0 ...`, but then when you read that you think, what is that `i0` doing there... My proposal was to have this in a module so that you can open it with `i0` and it is less confusing. An alternative would be to have a lemma `car-is-pointed` inside Lattice

fredrik-bakke: gotcha!

PR #321, opened 2026-09-19
chore: Get rid of "equiv to" pattern

fredrik-bakke:
Renames
- `is-single←equiv-to-single` to `is-single←equiv`
- `is-prop←equiv-to-prop` to `is-prop←equiv`
- `Universal←equiv-to-universal` to `←Universal←equiv`


fredrik-bakke:
Gonna pull a cheeky one and merge this one too without review

PR #314, opened 2026-09-17
Relative separations

fredrik-bakke:
Defines relative separations and shows they are reflective and accessible under appropriate conditions. Moreover, shows that relative separation at terminal projections is Sigma-closed, and relative separation along decidable embeddings are Sigma-closed, given that the underlying modality contains all propositions.

Builds on #312.

PR #322, opened 2026-09-19
𝑛-Truncated categories

fredrik-bakke:
Defines 𝑛-truncated categories and the homotopy 𝑛-category with its universal property.

Builds on #314.


fredrik-bakke:
I'm starting to think the choice to have simpliciality be part of the definition of categories is a bad one. Especially here because it means when we construct the homotopy n-category we also simplicially localise, which just doesn't sounds right in my ear.

fredrik-bakke:
I included this neat diagram of the generator map for hom 0-truncation :)
image

fredrik-bakke:
I changed the phrasing to "n-truncated categories", so we can reserve n-categories for the true notion, also known as $(\infty,n)$-categories.

fredrik-bakke:
Might want to write a changelog for this PR

PR #312, opened 2026-09-17
Relative suspensions

fredrik-bakke:
Formalises relative suspensions, a generalisation of suspensions and a special case of pushout products.Given a map $δ \colon X \to Y$ and a type $C$, then the $\delta$**-suspension** of $C$ is the pushout $(Y × C) \amalg_{X × C} X$.

I show that this construction is functorial, and using the fold property of pushouts I show that we recover suspension as a special case. This is work towards generalising separation reflective subuniverses.

PR #325, opened 2026-09-20
Transposing commuting squares along equivalences

fredrik-bakke:
A simple small refactor PR introducing a basic lemma for transposing commuting squares along equivalences. Pulled out of #312


samtoth:
Just FYI, the new tree wasn't renamed from this PR. (But we can fix it in #312)

samtoth:
We should definitely figure out some kind of automatic check for this

fredrik-bakke:
> We should definitely figure out some kind of automatic check for this

Yuppers

PR #326, opened 2026-09-20
chore: Some performance improvements

fredrik-bakke:
Opacifies some frequently used lemmas and simplifies two performance heavy proofs.


samtoth:
Very nice improvements :)

